PR #26, opened 2021-11-15
[NCL-6755] Create first unit test for Project List

DnsZhou:

PR #30, opened 2021-11-18
[NCL-6755] Create first unit test for Project List

DnsZhou:
Updated https://github.com/project-ncl/pnc-web-ui-react/pull/26 for merging code conflicts


package.json
```diff
@@ -50,6 +51,7 @@
     ]
   },
   "devDependencies": {
+    "@types/react-test-renderer": "^17.0.1",
```
matedo1, line 54: Hi @DnsZhou  I can see `@types/react-test-renderer` dependency in devDependencies and `react-test-renderer` dependency in dependencies section, it there some special reason?

DnsZhou: this was recommended by the npm when I missed @types/react-test-renderer for typescript. seems the one with @types is just for development that could enable type script to link the types.

PR #32, opened 2021-12-07
[NCL-6818] Unit test for Projects Page component

DnsZhou:


src/services/__mocks__/projectService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Gets all Projects.
+   *
+   * @returns Projects
+   */
+  public getProjects() {
+    return new Promise((resolve, rejects) => {
+      import('../../services/__mocks__/projects-mock.json').then((mockProjectsRequest) => {
+        resolve({ data: mockProjectsRequest });
```
matedo1, line 14: I guess there will be the same pattern (lines 12-16) often used also on the other places like `buildServices` etc., we can consider creating new util method (let's say `requestMock()`) for it so that it could look like:

```ts
public getProjects() {
   return requestMock('../../services/__mocks__/projects-mock.json');
}
```

src/services/__mocks__/projectService.ts
```diff
@@ -0,0 +1,43 @@
+import { rejects } from 'assert';
+
+class ProjectService {
+  path = '/projects';
```
matedo1, line 4: This should be removed?

src/components/ProjectsList/__tests__/ProjectsList.test.tsx
```diff
@@ -1,14 +1,15 @@
...
+
+jest.mock('../../../services/projectService');
 
 describe('display ProjectList component', () => {
   let mockProjectsRequest: any;
   let mockProjects: any;
 
   async function loadMocks() {
-    mockProjectsRequest = await import('./data/mock-projects-request.json');
+    mockProjectsRequest = await import('../../../services/__mocks__/projects-mock.json');
```
matedo1, line 12: Using `const mockProjectsRequest` and removing `let mockProjectsRequest: any;` should be enough.

Do we want to keep the same convention as on the other places? Then maybe `projectsRequestMock` and `projectsMock` would be better.

src/components/ProjectsPage/__tests__/ProjectsPage.test.tsx
```diff
@@ -1,13 +1,48 @@
-import { render } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
```
matedo1, line 1: I would prefer using `@testing-library/react` over `react-dom/test-utils` for `act`

React team recommends the same https://testing-library.com/docs/react-testing-library/api#act

src/components/ProjectsList/__tests__/ProjectsList.test.tsx
```diff
@@ -1,14 +1,15 @@
 import { ProjectsList } from '../ProjectsList';
 import { MemoryRouter } from 'react-router-dom';
 import { render, screen } from '@testing-library/react';
-import renderer from 'react-test-renderer';
+
+jest.mock('../../../services/projectService');
 
 describe('display ProjectList component', () => {
   let mockProjectsRequest: any;
   let mockProjects: any;
```
matedo1, line 9: In the future we could try to use more specific type and avoid `any`

src/services/__mocks__/projectService.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Gets a specific Project.
+   *
+   * @param id - ID of the Project
+   * @returns Project
+   */
+  public getProject(id: string) {
+    return null;
```
matedo1, line 26: let's use `throw new Error('Not implemented yet')`;

matedo1 (changes requested):

src/components/ProjectsList/__tests__/ProjectsList.test.tsx
DnsZhou: yup that make sense

src/services/__mocks__/projectService.ts
DnsZhou: will be solved in NCL-6855

matedo1 (approved):

PR #46, opened 2022-02-03
[NCL-6849] Update React Router to v6

jomrazek:
Migration from react router v5 to version v6:
- `<Switch>` components replaced with `<Routes>`
- Relative Route paths and no `exact`
- Route children/component changed to element as children is reserved for nested routes
- `useNavigate` instead of `useHistory `- useHistory was removed in version 6 and useNavigate should be used for navigation, we use history.location feature that was replaced with `useLocation`
sources I used for this migration:
https://reactrouter.com/docs/en/v6/upgrading/v5#upgrade-to-react-router-v6
https://reactrouter.com/docs/en/v6/api#uselocation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,12 +17,11 @@
         "@types/node": "^12.20.15",
         "@types/react": "^17.0.11",
         "@types/react-dom": "^17.0.7",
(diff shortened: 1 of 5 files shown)


jakubvanko (approved):
Looks good to me

PR #47, opened 2022-02-04
[NCL-6807] Update npm dependencies

jakubvanko:
There were no problems updating these dependencies.
Library react-scripts was split into a new Jira due to a bug that it contains regarding SVGs.


jakubvanko:
Need to add package-lock.json for every commit

PR #48, opened 2022-02-08
NCL-6928 Fix back button for pages with the pagination

matedo1:


jomrazek (approved):
Change is working and code looks ok to me. :+1:

PR #53, opened 2022-03-08
[NCL-6924] Create examples for individual components

jakubvanko:


DnsZhou (changes requested):
Hi Jakub, the demo page looks good to me and verified from my local environment. Just one suggestion that you may implement an unit test for the demo page and all you need to do is a rendering test, see productsPage as a reference: https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/ProductsPage/__tests__/ProductsPage.test.tsx

PR #55, opened 2022-03-10
[NCL-6752] Implement auth service

jomrazek:
You need to specify: `export REACT_APP_KEYCLOAK_URL=https://<deployment-url>.com`
Skeleton is there and working, still need some improvements that I noticed:
- get rid off env var and read keycloak url in html from configuration file
- without kinit aka in private window in protected route is user always unauthenticated and login process is started, it will do it automatically but screen flickers because of this
- methods in keycloak service are redundant really, you can and I guess will call static keycloak instance and call methods on it than creating instance of keycloak service to call them 

anyway, look at it guys and I'm looking forward your feedback

### DIFF
--- a/src/AppRoutes.tsx
+++ b/src/AppRoutes.tsx
@@ -17,6 +17,27 @@ import { ScmRepositoriesPage } from './components/ScmRepositoriesPage/ScmReposit
 import { VariablesPage } from './components/VariablesPage/VariablesPage';
 import { NotFoundPage } from './components/NotFoundPage/NotFoundPage';
 import { AdministrationPage } from './components/AdministrationPage/AdministrationPage';
+import { keycloakService, AUTH_ROLE } from './services/keycloakService';
+
+interface IProtectedRouteProps {
+  role: AUTH_ROLE;
+}
+
+const ProtectedRoute = ({ children, role }: React.PropsWithChildren<IProtectedRouteProps>) => {
+  if (keycloakService.isAuthenticated()) {
+    if (keycloakService.hasRealmRole(role)) {
+      return <>{children}</>;
+    } else {
+      return <div>User not allowed</div>;
+    }
+  } else {
+    keycloakService.login().catch(() => {
+      throw new Error('Keycloak login failed.');
+    });
+  }
+
+  return <div>Redirecting to keycloak...</div>;
+};
 
 export const AppRoutes = () => (
   <Routes>
@@ -33,7 +54,14 @@ export const AppRoutes = () => (
 
     {/* special pages */}
     <Route path="admin/variables" element={<VariablesPage />} />
-    <Route path="admin/administration" element={<AdministrationPage />} />
+    <Route
+      path="admin/administration"
+      element={
+        <ProtectedRoute role={AUTH_ROLE.Admin}>
+          <AdministrationPage />
+        </ProtectedRoute>
+      }
+    />
     <Route path="*" element={<NotFoundPage />} />
   </Routes>
 );
--- a/src/__tests__/AppLayout.test.tsx
+++ b/src/__tests__/AppLayout.test.tsx
@@ -2,9 +2,16 @@ import { render } from '@testing-library/react';
 import { AppLayout } from '../AppLayout';
 import { MemoryRouter } from 'react-router-dom';
 
+jest.mock('../services/keycloakService');
+
 window.pnc = {
   config: {
     userGuideUrl: 'https://localhost:3000/',
+    keycloak: {
+      url: 'https://localhost:3000/',
+      realm: 'test',
+      clientId: 'test',
+    },
   },
 };
 
--- a/src/services/__mocks__/keycloakService.ts
+++ b/src/services/__mocks__/keycloakService.ts
@@ -1,0 +1,50 @@
+class KeycloakServiceMock {
+  private initialized: boolean = false;
+  private user: any;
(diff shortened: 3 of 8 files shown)


src/AppRoutes.tsx
```diff
@@ -17,6 +17,11 @@ import { ScmRepositoriesPage } from './components/ScmRepositoriesPage/ScmReposit
 import { VariablesPage } from './components/VariablesPage/VariablesPage';
 import { NotFoundPage } from './components/NotFoundPage/NotFoundPage';
 import { AdministrationPage } from './components/AdministrationPage/AdministrationPage';
+import { keycloakService, AUTH_ROLE } from './services/keycloakService';
+
```
matedo1, line 21: This file should be renamed from `AppRoutes.tsx` to `Routes.tsx`

src/services/keycloakService.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    this.keycloakAuth = window.Keycloak({
+      url: keycloakConfig.url,
+      realm: keycloakConfig.realm,
+      clientId: keycloakConfig.clientId,
+    });
+
+    return new Promise((resolve, reject) => {
+      this.keycloakAuth
+        .init({ onLoad: 'check-sso', checkLoginIframeInterval: 1 })
```
matedo1, line 47: Default value for `checkLoginIframeInterval` is `5` ( https://github.com/keycloak/keycloak-documentation/blob/main/securing_apps/topics/oidc/javascript-adapter.adoc ), is there some reason to change it? If not, I would use the default value.

src/services/keycloakService.ts
```diff
@@ -0,0 +1,122 @@
+import * as WebConfigAPI from '../services/WebConfigService';
+declare global {
+  interface Window {
+    Keycloak?: any;
+  }
+}
+/**
+ * Enum with possible authentication roles in keycloak
```
matedo1, line 8: Please use `.` at the end, so `...on roles in keycloak.` See for example `useDataContainer.ts` file as an example. The same on other similar places, I remember having some issues when auto generating docs while missing it.

src/services/keycloakService.ts
```diff
@@ -0,0 +1,122 @@
...
+   * @returns String
+   */
+  public getToken(): String {
+    return this.keycloakAuth.token;
+  }
+
+  /**
+   * Gets user name from keycloak
+   *
+   * @returns String
```
matedo1, line 103: what value is returned when the user is not logged in? `returns` should be focused on values. Otherwise other developers can not be sure AppLayout.tsx:71 (and other occurrences) will be evaluated properly.

src/services/keycloakService.ts
```diff
@@ -0,0 +1,122 @@
...
+   * @param redirectUri uri to redirect after logout
+   */
+  public logout(redirectUri?: String): void {
+    this.keycloakAuth.logout({ redirectUri });
+  }
+
+  /**
+   * Gets keycloak token
+   *
+   * @returns String
```
matedo1, line 94: what value is returned when the user is not logged in? `returns` should be focused on values.

src/AppRoutes.tsx
```diff
@@ -33,7 +38,30 @@ export const AppRoutes = () => (
...
+    } else {
+      return <div>User not allowed</div>;
+    }
+  } else {
+    keycloakService.login().catch(() => {
+      throw new Error('Keycloak login failed.');
+    });
+  }
+
+  return <NotFoundPage />;
```
matedo1, line 66: I think some error should be reported instead of displaying `NotFoundPage`? Let's look at it during demo meeting.

matedo1 (changes requested):
Just a few minor changes. Looking good, thanks!

matedo1:
Hi @DnsZhou , can you check test related parts?

src/services/__mocks__/keycloakService.ts
```diff
@@ -0,0 +1,9 @@
+class KeycloakServiceMock {
```
DnsZhou, line 1: More functions may need to be implemented to proceed with more tests for the login and logout operations.

src/__tests__/AppLayout.test.tsx
```diff
@@ -2,9 +2,16 @@ import { render } from '@testing-library/react';
...
 window.pnc = {
   config: {
     userGuideUrl: 'https://localhost:3000/',
+    keycloak: {
+      url: 'https://localhost:3000/',
+      realm: 'test',
+      clientId: 'test',
+    },
   },
 };
```
DnsZhou, line 17: Since now we have several interactions with the front end about the login/user operation, you can implement some unit tests to mock the login and log out process, for example, to verify whether the user name is shown properly as the response from keycloak passes, or whether logout process can proceed successfully and user can log in again with the new user identity. You can use the mock keycloak service you added.

DnsZhou (changes requested):
Hi Jozef, thanks for the commit there are just some missing unit tests that need to go along with the login feature as well. See specific comments I left and feel free to discuss with us about that. Thanks! @jomrazek

jomrazek:
@DnsZhou Thanks for your input Dennis, I extended keycloak mock hopefully to somewhat working mock of full service, do you think we can move creation of new test to new issue as I don't have much knowledge about this and would like to research it a bit first.

DnsZhou (approved):

PR #58, opened 2022-03-22
[NCL 6583] Implement ProjectDetailPage

jakubvanko:


jakubvanko:
Hi @DnsZhou could you please advise me on the best way to create unit tests for this page? My naive way is not working properly.

src/AppRoutes.tsx
```diff
@@ -46,6 +47,7 @@ export const AppRoutes = () => (
     {/* entity pages */}
     <Route path="products" element={<ProductsPage />} />
     <Route path="projects" element={<ProjectsPage />} />
+    <Route path="projects/:id" element={<ProjectDetailPage />} />
```
matedo1, line 50: @jakubvanko @jomrazek WDYT about using nested routes in cases like this? https://reactrouter.com/docs/en/v6/getting-started/concepts#nested-routes

matedo1, line 50: @jakubvanko @jomrazek WDYT about using nested routes in cases like this? https://reactrouter.com/docs/en/v6/getting-started/concepts#nested-routes

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  create: <FileIcon />,
+  edit: <PencilAltIcon />,
+  delete: <TrashIcon />,
+};
+
+interface IActionButton {
+  actionType: 'create' | 'edit' | 'delete';
+}
+
+export const ActionButton = ({ actionType }: IActionButton) => {
```
matedo1, line 16: let's use shorter syntax in cases like this, `{ return` part can be skipped, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/EmptyStates/EmptyStateCard.tsx#L13

matedo1, line 16: let's use shorter syntax in cases like this, `{ return` part can be skipped, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/EmptyStates/EmptyStateCard.tsx#L13

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  );
+};
+
+interface IActionHeader extends IActionButton {
+  text: string;
+}
+
+export const ActionHeader = ({ text, ...actionButtonProps }: IActionHeader) => {
+  return (
+    <Flex justifyContent={{ default: 'justifyContentSpaceBetween' }}>
```
matedo1, line 30: Check unintentional re-renders https://www.debugbear.com/blog/react-rerenders#passing-objects-as-props

matedo1, line 30: Check unintentional re-renders https://www.debugbear.com/blog/react-rerenders#passing-objects-as-props , there are some other cases, please fix them too.

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import styles from './ActionButton.module.css';
+
+const actionIcons = {
+  create: <FileIcon />,
+  edit: <PencilAltIcon />,
+  delete: <TrashIcon />,
+};
+
+interface IActionButton {
```
matedo1, line 12: In the end we decided to change naming guidelines a little bit, please use `IActionButtonProps` instead (the same is valid for all other places)

matedo1, line 12: In the end we decided to change naming guidelines a little bit, please use `IActionButtonProps` instead (the same is valid for all other places)

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
+import { Button, Flex } from '@patternfly/react-core';
```
matedo1, line 1: This file is named like `ActionButton.tsx` and it exports two components:
1) ActionButton
2) ActionHeader

Which one is considered to be the default one that should be commonly used? Maybe creating new file for each component would be better? Or removing `export` from `ActionButton` and renaming this file to `ActionHeader.tsx`? Let's discuss it.

matedo1, line 1: This file is named like `ActionButton.tsx` and it exports two components:
1) ActionButton
2) ActionHeader

Which one is considered to be the default one that should be commonly used? Maybe creating new file for each component would be better? Or removing `export` from `ActionButton` and renaming this file to `ActionHeader.tsx`? Let's discuss it.

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,63 @@
+import { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import { Grid, Card, CardTitle, CardBody, FlexItem, Flex } from '@patternfly/react-core';
+import { PageLayout } from './../PageLayout/PageLayout';
+import { useDataContainer } from '../../containers/DataContainer/useDataContainer';
+import { projectService } from '../../services/projectService';
+import { DataContainer } from '../../containers/DataContainer/DataContainer';
+import { ActionHeader } from '../ActionButton/ActionButton';
+
+const ProjectAttribute = ({ attribute, value }: { attribute: string; value: string }) => {
```
matedo1, line 10: let's use shorter syntax in cases like this, `{ return` part can be skipped, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/EmptyStates/EmptyStateCard.tsx#L13

matedo1, line 10: let's use shorter syntax in cases like this, `{ return` part can be skipped, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/EmptyStates/EmptyStateCard.tsx#L13

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { id } = useParams();
+
+  const dataContainer = useDataContainer((requestConfig: Object) => projectService.getProject(requestConfig, id!));
+
+  // TODO: Create a better solution than disabling the next line
+  useEffect(() => dataContainer.refresh({ id }), []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  return (
+    <DataContainer {...dataContainer} title="Project Details">
+      <PageLayout title={dataContainer.data?.name} description={dataContainer.data?.description}>
```
matedo1, line 29: `data?` using `?` should not be needed, `data` should be always valid, see https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/containers/DataContainer/DataContainer.tsx#L46

matedo1, line 29: `data?` using `?` should not be needed, `data` should be always valid, see https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/containers/DataContainer/DataContainer.tsx#L46

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+      {actionType}
+    </Button>
+  );
+};
+
+interface IActionHeader extends IActionButton {
+  text: string;
+}
+
+export const ActionHeader = ({ text, ...actionButtonProps }: IActionHeader) => {
```
matedo1, line 28: Please add documentation.

matedo1, line 28: Please add documentation.

src/components/ProjectDetailPage/ProjectDetailPage.tsx
matedo1, line 10: I think it makes sense to create more general component (in a new file), the same "issue" needs to be solved on all other places and we need to use the same solution to be consistent.

matedo1, line 10: I think it makes sense to create more general component, the same "issue" needs to be solved on all other places.

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+  );
+};
+
+export const ProjectDetailPage = () => {
+  const { id } = useParams();
+
+  const dataContainer = useDataContainer((requestConfig: Object) => projectService.getProject(requestConfig, id!));
+
+  // TODO: Create a better solution than disabling the next line
+  useEffect(() => dataContainer.refresh({ id }), []); // eslint-disable-line react-hooks/exhaustive-deps
```
matedo1, line 25: Why is passing `{ id }` needed?

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Flex justifyContent={{ default: 'justifyContentSpaceBetween' }}>
+      {text} <ActionButton {...actionButtonProps} />
```
matedo1, line 31: Why `...` syntax is used? What happened with other parameters different from `actionType`?

matedo1, line 31: Why `...` syntax is used? What happened with other parameters different from `actionType`?

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+                  <ProjectAttribute attribute="Technical Leader" value={dataContainer.data?.technicalLeader} />
+                </Grid>
+              </CardBody>
+            </Card>
+          </FlexItem>
+
+          <FlexItem>
+            <Card>
+              <CardTitle>
+                <ActionHeader actionType="create" text="Build Configs" />
```
matedo1, line 50: Please move `Build Configs` title outside of the white box, to make it similar to our PM discussion (`Mar 16, 2:28 PM`)

matedo1, line 50: Let's move `Build Configs` title outside of the white box, to make it similar to our PM discussion (`Mar 16, 2:28 PM`)

src/components/ProjectsList/ProjectsList.tsx
```diff
@@ -23,7 +23,7 @@ export const ProjectsList = ({ projects }: IProjectsList) => {
 
   const rows = projects.map((project: Project) => [
     {
-      title: <Link to="#">{project.name}</Link>,
+      title: <Link to={project.id}>{project.name}</Link>,
```
matedo1, line 26: Let's create new component, something like `ProjectLink` with all required attributes, in this case it would be only `id: string` and then use it here

matedo1, line 26: Let's create new component, something like `ProjectLink` with all required attributes, in this case it would be only `id: string`

src/services/projectService.ts
```diff
@@ -18,8 +18,8 @@ class ProjectService {
    * @param id - ID of the Project
    * @returns Project
    */
-  public getProject(id: string) {
-    return pncClient.getHttpClient().get(`${this.path}/${id}`);
+  public getProject(requestConfig = {}, id: string) {
```
matedo1, line 21: let's swap arguments to: `getProject(id: string, requestConfig = {})`, and WDYT about using named arguments?

matedo1, line 21: let's swap arguments to: `getProject(id: string, requestConfig = {})`, and WDYT about using named arguments?

src/components/ActionButton/ActionButton.tsx
matedo1: What `text` is used for? What kind of information is trying to display? Please add documentation.

matedo1: What `text` is used for? What kind of information is trying to display? Please add documentation.

src/AppRoutes.tsx
jomrazek: I believe this is exact place to use them :+1:

src/components/ProjectDetailPage/__tests__/ProjectDetailPage.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+test('renders ProjectDetailPage', () => {
+  render(
+    <MemoryRouter initialEntries={['projects/101']}>
+      <Route path="projects/:id">
+        <ProjectDetailPage />
+      </Route>
+    </MemoryRouter>
+  );
+});
```
DnsZhou, line 13: Hi @jakubvanko ,you can refer to Unit Test for ProjectListPage to have similar test for your page, see https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/ProjectsList/__tests__/ProjectsList.test.tsx
For example, you can test for "rendering ProjectDetails page to have the right data" and also have a snopshot test like that proceeded in ProjectsList.test.tsx.

DnsZhou: e.g, to verify some data exist on the page ma looks like this:
`    const firstProject = screen.getByText(projectsMock[0].name);
    expect(firstProject).toBeInTheDocument();`

and code for verify the snapshot can be found here: https://github.com/project-ncl/pnc-web-ui-react/blob/f5e16df28ff5ddfe8105dbcf5236349cdd8be008/src/components/ProjectsList/__tests__/ProjectsList.test.tsx#L33-L40

DnsZhou (changes requested):
Hi @jakubvanko , thanks for the work on Project Detail Page! you can refer to the comments on `/__tests__/ProjectDetailPage.test.tsx`to finish the unit test part.

src/components/ActionButton/ActionButton.tsx
jakubvanko: ActionButton is only the main button component whereas the action header uses the button to hide some more logic (so it doesnt have to repeat). I think both may be used and it may be difficult to decide which is the primary one. I will create a new file because these components may get more complex in the future.

src/AppRoutes.tsx
jakubvanko: Personally, I think that we should update all of the paths at once since there are other candidates for such a change (we already have admin pages admin/demo, admin/variables, admin/administration) and I believe that mixing two approaches could be confusing.

src/components/ActionButton/ActionButton.tsx
jakubvanko: Decided to replace text with children and remove the spread operator

src/components/ProjectDetailPage/ProjectDetailPage.tsx
jakubvanko: Created a new NameValueText component

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+              <CardTitle>
+                <ActionHeader actionType="edit">Attributes</ActionHeader>
+              </CardTitle>
+              <CardBody>
+                <Grid span={5} hasGutter component={'span'}>
+                  <NameValueText name="Project URL" value={dataContainer.data?.projectUrl} />
+                  <NameValueText name="Issue Tracker URL" value={dataContainer.data?.issueTrackerUrl} />
+                  <NameValueText name="Engineering Team" value={dataContainer.data?.engineeringTeam} />
+                  <NameValueText name="Technical Leader" value={dataContainer.data?.technicalLeader} />
+                </Grid>
```
jakubvanko, line 36: The layout of flex and grid could be perhaps improved.

src/AppRoutes.tsx
matedo1: @jakubvanko yes, all paths will be already udpated (it's part of the PR 62), the best will be wait once PR 62 is merged and rebase this one on top of it

jakubvanko: Okay, will change it once the PR62 is merged

src/components/ProjectDetailPage/ProjectDetailPage.tsx
jakubvanko: Removed

src/components/ProjectDetailPage/ProjectDetailPage.tsx
jakubvanko: Was reworked into KeyValueTable

src/components/ProjectDetailPage/ProjectDetailPage.tsx
jakubvanko: @matedo1  Also, if we decide to move forward with the design, perhaps we could think about generalizing the page so that its layout could be reused.

jakubvanko:
Hi @DnsZhou I added the unit tests. Could you please have a look whether it is okay now? Thank you.

jakubvanko:
- Fix object props: const flexDirection: FlexProps['direction'] = { default: 'column' }; (outside of component)
- KeyValueTable to accept also components
- JIRA for eslint warning
- ActionHeader for more or zero buttons

src/components/KeyValueTable/KeyValueTable.tsx
```diff
@@ -0,0 +1,37 @@
+import { Grid, GridItem } from '@patternfly/react-core';
+
+import styles from './KeyValueTable.module.css';
+
+interface IKeyValueEntryProps {
+  name: React.ReactNode;
+  value: React.ReactNode;
+}
+
+const KeyValueEntry = ({ name, value }: IKeyValueEntryProps) => (
```
matedo1, line 10: let's use `AttributesItem` + the same rule for interface

src/components/KeyValueTable/KeyValueTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+    </GridItem>
+    <GridItem md={9} sm={12} className={value ? '' : styles['value']}>
+      {value ?? 'Empty'}
+    </GridItem>
+  </>
+);
+
+interface IKeyValueTableProps {
+  keyValueArray: {
+    key: React.ReactNode;
```
matedo1, line 23: let's use `name` to be consistent with the component above + to avoid `key` conflicts ( https://reactjs.org/docs/lists-and-keys.html )

src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -0,0 +1,24 @@
...
+const flexJustifyContent: FlexProps['justifyContent'] = { default: 'justifyContentSpaceBetween' };
+const flexAlignItems: FlexProps['alignItems'] = { default: 'alignItemsCenter' };
+
+/**
+ * Layout component for section headers.
+ *
+ * @param children - the content of the (textual) heading
+ * @param sideComponent - the content on the opposite (right) side of the main heading component
+ */
+export const SectionHeader = ({ children, sideComponent }: ISectionHeaderProps) => (
```
matedo1, line 17: let's use `actions` instead of `sideComponent`

src/components/ActionButton/ActionButton.module.css
```diff
@@ -0,0 +1,4 @@
+.action-button {
+  width: 100px;
```
matedo1, line 2: Width like this is too hardcoded, let's use something like padding-left and padding-right

src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -0,0 +1,24 @@
...
+  <Flex justifyContent={flexJustifyContent} alignItems={flexAlignItems} className="m-b-10">
+    <TextContent>
+      <Text component={TextVariants.h2}>{children}</Text>
```
matedo1, line 20: prevent generating empty elements when no `children` is passed in, but also make sure there is no top margin missing after it's removed

matedo1 (changes requested):
Thanks for this PR, it's very nice one! There are just minor changes to be applied

src/components/ProjectDetailPage/__tests__/ProjectDetailPage.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ProjectDetailPage } from '../ProjectDetailPage';
+import { MemoryRouter } from 'react-router-dom';
+
+jest.mock('../../../services/projectService');
+
+describe('display ProjectDetailPage component', () => {
+  test('renders ProjectDetail', () => {
+    render(
+      <MemoryRouter>
+        <ProjectDetailPage />
```
DnsZhou, line 11: Hi Jakub,
Since you have already gotten the mock data from mockProjectService, you should also verify that the data were displayed properly by getting some element or text from the screen to see if it exists or match with your mock data. See my last review about how to implement some tests over the elements on the page.

DnsZhou (changes requested):
Hi Jakub, seems you missed the verification about whether the data displayed on the page match with the mock service data. See the specific comments and feel free to reach me if you have any questions.

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+import styles from './ActionButton.module.css';
+
+const actionIcons = {
+  create: <FileIcon />,
+  edit: <PencilAltIcon />,
+  delete: <TrashIcon />,
+};
+
+export interface IActionButtonProps {
+  actionType: 'create' | 'edit' | 'delete';
```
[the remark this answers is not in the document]
jakubvanko: New JIRA = NCL-7050

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const dataContainer = useDataContainer(({ requestConfig }: IService) =>
+    projectService.getProject({ id: projectId as string }, requestConfig)
+  );
+
+  // TODO: Create a better solution than disabling the next line
+  useEffect(() => dataContainer.refresh({ requestConfig: {} }), []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const attributes = [
+    {
+      key: 'Project URL',
```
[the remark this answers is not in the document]
jakubvanko: Added to demo page

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Represents a button component with a predefined icon.
+ * Can also serve as a Link (if link prop is specified).
+ *
+ * @param actionType - specifies the icon of the button, possible options are 'create' | 'edit' | 'delete'
+ * @param link - optional prop if the button should serve as a Link component (will redirect to the specified link)
+ * @param onClick - function to perform on clicking the button
+ */
+export const ActionButton = ({ actionType, link, onClick }: IActionButtonProps) => (
+  <ConditionalLinkWrapper link={link}>
+    <Button variant="secondary" isSmall className={styles['action-button']} icon={actionIcons[actionType]} onClick={onClick}>
```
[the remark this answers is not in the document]
jakubvanko: Nothing should happen, it is an optional argument

src/components/ActionButton/ActionButton.module.css
jakubvanko: Added padding and changed it to min-width. If the option is not set, the buttons will have different sizes, is this expected?

DnsZhou (approved):

DnsZhou:
Hi Kubo, it Looks good from my side, and verified it in my local environment. You can reach Martin to see if there is anyting else that needs to be changed. Thx.

PR #70, opened 2022-05-31
[NCL-7050] Improve ActionButton to make it more generic

jakubvanko:


src/components/ActionButton/ActionButton.tsx
```diff
@@ -1,42 +1,51 @@
...
+  quality: <TagIcon />,
+  external: <ExternalLinkAltIcon />,
+  close: <LockIcon />,
+  mark: <FlagIcon />,
+} as const;
 
 export interface IActionButtonProps {
-  actionType: 'create' | 'edit' | 'delete';
+  iconType: keyof typeof iconDictionary;
+  children?: React.ReactNode;
```
matedo1, line 27: use `React.PropsWithChildren` instead, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/EmptyStates/StateCard.tsx#L18

src/components/ProjectDetailPage/ProjectDetailPage.tsx
```diff
@@ -43,15 +43,27 @@ export const ProjectDetailPage = () => {
...
           <Card>
             <CardBody>
               <AttributesItems attributes={attributes} />
             </CardBody>
           </Card>
         </div>
-        <SectionHeader actions={<ActionButton action={() => console.log('Not implemented yet!')} actionType="create" />}>
+        <SectionHeader
+          actions={
+            <ActionButton action={() => console.log('Not implemented yet!')} iconType="create">
```
matedo1, line 62: Actually Project Create page is already implemented: `/projects/create`

matedo1 (changes requested):

src/components/ProjectDetailPage/ProjectDetailPage.tsx
jakubvanko: I think we need to use a link to a BuildConfig create page tho.

matedo1: you are right! Deleting this comment, as it's not valid

matedo1 (commented):
Could you try to add also support for a button without the icon?

PR #82, opened 2022-07-28
[NCL-7243] Add validation to input fields on Project pages (first validation implementation)

patrikk0123:
Backend errors are still displayed on form-level.
Validation backend REST API endpoints needed.
Also would be nice if error server reponse included more info (for example input field id).

### DIFF
--- a/src/components/DemoPage/DemoPage.tsx
+++ b/src/components/DemoPage/DemoPage.tsx
@@ -30,7 +30,7 @@ import { InfoCircleIcon } from '@patternfly/react-icons';
 import { ActionButton } from '../ActionButton/ActionButton';
 import mockBuildData from './data/mock-build-data.json';
 import { useTitle } from '../../containers/useTitle';
-import { useForm } from '../../containers/useForm';
+import { IFields, useForm } from '../../containers/useForm';
 import { minLength, maxLength } from '../../utils/formValidationHelpers';
 
 const buildRes: Build[] = mockBuildData;
@@ -59,9 +59,9 @@ const formConfig = {
 export const DemoPage = () => {
   useTitle('Demo Page');
 
-  const submitForm = () => {
+  const submitForm = (data: IFields) => {
     console.log('form state when submitted:', {
-      ...fields,
+      ...data,
     });
   };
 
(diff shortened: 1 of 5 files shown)


matedo1 (changes requested):

src/utils/formValidationHelpers.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Creator of a min length validation function.
+ *
+ * Returned function accepts empty string.
+ *
+ * @param min - Minimal length of a string returned function will validate
+ * @returns function validating minimal length
+ */
+export const minLength = (min: number): Function => {
```
matedo1, line 28: please add some demo on Demo page, the same for `maxLength`

matedo1 (changes requested):

src/containers/useForm.ts
```diff
@@ -0,0 +1,208 @@
...
+      state: 'default' as TextInputProps['validated'],
+    };
+    validate(newFieldState);
+    setFields({ ...fields, [fieldName]: newFieldState });
+
+    setHasChanged(true);
+  };
+
+  // validate field state and change error messages / state
+  const validate = (fieldState: IField) => {
```
matedo1, line 130: let's use only `field` instead of `fieldState` everywhere

Even if `addError` does the check for empty error, I would call this method only if you know you want to add new error

```js
const validate = (fieldState: IField) => {
  if (fieldState.isRequired) {
    if (!fieldState.value) {
      addError(fieldState, 'Field must be filled.');
    }
    setState(fieldState);
  }
  if (fieldState.validators) {
    for (const validator of fieldState.validators) {
      if (!validator.validator(fieldState.value)) {
        addError(fieldState, validator.errorMessage);
      }
    }
    setState(fieldState);
  }
};
```

matedo1 (changes requested):
Just minor changes, you can also update TSDoc, thanks for implementing this!

src/components/DemoPage/DemoPage.tsx
```diff
@@ -12,14 +30,160 @@ import { InfoCircleIcon } from '@patternfly/react-icons';
...
+    value: '',
+    isRequired: true,
+  },
+};
+
 export const DemoPage = () => {
-  useTitle('Demo');
+  useTitle('Demo Page');
+
+  const submitForm = () => {
```
matedo1, line 62: data attribute should be passed?

matedo1 (changes requested):

matedo1 (approved):

PR #84, opened 2022-08-10
[NCL-7256] Keep UI partially working even if KeyCloak is not available

patrikk0123:


matedo1:
Hi @jomrazek , could you please do a review for @patrikk0123 ?

jomrazek (commented):
looks ok to me

src/AppRoutes.tsx
```diff
@@ -18,20 +18,25 @@ import { ProjectDetailPage } from './components/ProjectDetailPage/ProjectDetailP
 // special pages
 import { DemoPage } from './components/DemoPage/DemoPage';
 import { VariablesPage } from './components/VariablesPage/VariablesPage';
-import { NotFoundPage } from './components/NotFoundPage/NotFoundPage';
 import { AdministrationPage } from './components/AdministrationPage/AdministrationPage';
 import { keycloakService, AUTH_ROLE } from './services/keycloakService';
+import { ErrorPage } from './components/ErrorPage/ErrorPage';
 
 interface IProtectedRouteProps {
+  title?: string;
```
matedo1, line 26: I would be strict in this case for now, let's remove `?` character

src/AppRoutes.tsx
```diff
@@ -81,12 +86,12 @@ export const AppRoutes = () => (
...
         element={
-          <ProtectedRoute role={AUTH_ROLE.Admin}>
+          <ProtectedRoute title="Administration" role={AUTH_ROLE.Admin}>
             <AdministrationPage />
           </ProtectedRoute>
         }
       />
     </Route>
-    <Route path="*" element={<NotFoundPage />} />
+    <Route path="*" element={<ErrorPage errorDescription="The requested resource could not be found."></ErrorPage>} />
```
matedo1, line 95: I would use also `title` attribute `Page not found`, but define it inside https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/utils/PageTitles.ts like `pageNotFound`

src/components/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,23 @@
+import { CubesIcon } from '@patternfly/react-icons';
+import { Link } from 'react-router-dom';
+import { StateCard } from '../EmptyStates/StateCard';
+import { PageLayout } from '../PageLayout/PageLayout';
+import styles from './ErrorPage.module.css';
+
+interface IErrorPage {
+  pageTitle?: string;
```
matedo1, line 8: I would be strict also in this case, let's remove `?`, then also default values bellow will be removed

src/components/ErrorPage/ErrorPage.module.css
```diff
@@ -0,0 +1,3 @@
+.return-button-margin {
+  margin-top: 15px;
```
matedo1, line 2: for generic margins, paddings, etc, we use so called generics (see index.css), let's use `m-t-15: 15px` and remove this css module

src/AppRoutes.tsx
```diff
@@ -52,15 +57,15 @@ export const AppRoutes = () => (
       <Route
         path="create"
         element={
-          <ProtectedRoute>
+          <ProtectedRoute title="Create Project">
```
matedo1, line 60: let's define all titles inside https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/utils/PageTitles.ts

matedo1 (changes requested):

src/services/pncClient.ts
```diff
@@ -26,7 +26,7 @@ class PncClient {
     httpClient.interceptors.request.use((config) => {
       // perform actions before request is sent
 
-      if (keycloakService.isAuthenticated()) {
+      if (keycloakService.isKeycloakAvailable && keycloakService.isAuthenticated()) {
```
jomrazek, line 29: if we care just about true or false, can't we move isKeycloakAvailable check into service? Same for getUser and hasRealmRole in appLayout? Seems like we would need to do this check in all classes when using those methods while they can return false from service when we don't need to differentiate the failure point.

patrikk0123:
Changes had to be done, continuation here:
https://github.com/project-ncl/pnc-web-ui-react/pull/87

PR #85, opened 2022-08-10
[NCL-7275] Use import sort plugin for Prettier

patrikk0123:
For now, trivago's prettier sort imports plugin was installed and Prettier config was changed to use it.
https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports

It should be decided which order suits best and then applied to all files.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,15 @@
 {
   "singleQuote": true,
   "trailingComma": "es5",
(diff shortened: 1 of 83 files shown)


matedo1:
Hi @patrikk0123 , one more commit fixing all broken files needs to be added

matedo1 (approved):

PR #86, opened 2022-08-11
[NCL-6964] Create dependency Tree component

DnsZhou:


src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { useState, useEffect, useCallback } from 'react';
+import { DataContainer } from '../../containers/DataContainer/DataContainer';
+import { IService, useDataContainer } from '../../containers/DataContainer/useDataContainer';
+import { buildService } from '../../services/buildService';
+import { groupBuildService } from '../../services/groupBuildService';
+import { isBuild } from '../../utils/entityRecognition';
+import styles from './DependencyTree.module.css';
+import { BuildStatus } from '../BuildStatus/BuildStatus';
+
+export interface IGraphEdge {
```
matedo1, line 13: avoid `export` if it's not needed

src/services/buildService.ts
```diff
@@ -24,6 +24,16 @@ class BuildService {
   public getBuildCount(requestConfig: AxiosRequestConfig = {}) {
     return pncClient.getHttpClient().get(`${this.path}/count`, requestConfig);
   }
+
+  /**
+   * Gets dependency graph for a build.
+   * @returns DependencyGraph
```
matedo1, line 30: what does it return if id is falsy value?

src/services/buildService.ts
```diff
@@ -24,6 +24,16 @@ class BuildService {
...
+   */
+  public getDependencyGraph(id: string, requestConfig: AxiosRequestConfig = {}) {
+    if (id) {
```
matedo1, line 33: if `id` is falsy, it's developer mistake? if yes, some error should be fired?

src/services/groupBuildService.ts
```diff
@@ -0,0 +1,19 @@
...
+
+class GroupBuildService {
+  path = '/group-builds';
+
+  /**
+   * Gets dependency graph for a build.
+   * @returns DependencyGraph
+   */
+  public getDependencyGraph(id: string, requestConfig: AxiosRequestConfig = {}) {
+    return pncClient.getHttpClient().get(`${this.path}/${id}/dependency-graph`, requestConfig);
```
matedo1, line 12: `id` is not checked here, but it's checked in `buildService`, why?

src/services/groupBuildService.ts
```diff
@@ -0,0 +1,19 @@
+import { AxiosRequestConfig } from 'axios';
+import { pncClient } from './pncClient';
+
+class GroupBuildService {
+  path = '/group-builds';
+
+  /**
+   * Gets dependency graph for a build.
```
matedo1, line 8: group build

matedo1 (changes requested):

src/services/buildService.ts
DnsZhou: I removed the restriction to null id because it was not removed from other services like project service. also either a invalid format of id or id that has no data will be handled by data container.
image
image

src/services/buildService.ts
DnsZhou: image

src/services/buildService.ts
```diff
@@ -2,6 +2,10 @@ import { AxiosRequestConfig } from 'axios';
 import { kafkaClient } from './kafkaClient';
 import { pncClient } from './pncClient';
 
+export interface IBuildServiceData {
```
matedo1, line 5: `export` is not needed at this moment

src/services/groupBuildService.ts
```diff
@@ -1,14 +1,20 @@
 import { AxiosRequestConfig } from 'axios';
 import { pncClient } from './pncClient';
 
+export interface IGroupBuildServiceData {
```
matedo1, line 4: `export` is not needed at this moment

src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,262 @@
...
+        return buildService.getDependencyGraph({ id: serviceData!.build!.id }, requestConfig);
+      } else {
+        return groupBuildService.getDependencyGraph({ id: serviceData!.groupBuild!.id }, requestConfig);
+      }
+    }, [])
+  );
+
+  const refreshDataContainer = dataContainer.refresh;
+
+  const refreshComponent = () => {
```
matedo1, line 73: isn't possible to use just `useEffect` or some other way? This implementation looks like workaround a little bit

src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,262 @@
...
+      return rootNodes;
+    };
+
+    const attachChildFromEdges = (
+      currentNode: IDependencyDataItem,
+      edgesData: Array<IGraphEdge>,
+      nodesData: Map<string, IDependencyBuild>,
+      level: number
+    ) => {
+      level = level ? level : 1;
```
matedo1, line 148: can you use default parameter ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters ) instead?

src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { isBuild } from '../../utils/entityRecognition';
+import styles from './DependencyTree.module.css';
+import { BuildStatus } from '../BuildStatus/BuildStatus';
+
+interface IGraphEdge {
+  source: string;
+  target: string;
+}
+
+interface IGraphVertic {
```
patrikk0123, line 18: Should be IGraphVertex.

src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+    const attachChildFromEdges = (
+      currentNode: IDependencyDataItem,
+      edgesData: Array<IGraphEdge>,
+      nodesData: Map<string, IDependencyBuild>,
+      level: number
+    ) => {
+      level = level ? level : 1;
+      if (level === 1 && !isBuild(buildItem)) {
+        const rootNodes: Array<string> = level === 1 ? getRootNodes(edgesData, nodesData) : [];
```
patrikk0123, line 150: no point using ternary operator - level will always be 1, since it is required by if condition

patrikk0123 (dismissed):

src/components/DependencyTree/DependencyTree.tsx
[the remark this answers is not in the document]
DnsZhou: The level can be other than 1. this is a recursive function so it will pass level+1 for every child of it. see https://github.com/DnsZhou/pnc-web-ui-react/blob/d3e71c56fc9b00c291355ff8c5e0ed2577543394/src/components/DependencyTree/DependencyTree.tsx#L186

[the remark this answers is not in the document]
DnsZhou: > no point using ternary operator - level will always be 1, since it is required by if condition

The level can be other than 1. this is a recursive function so it will pass level+1 for every child of it. see https://github.com/DnsZhou/pnc-web-ui-react/blob/d3e71c56fc9b00c291355ff8c5e0ed2577543394/src/components/DependencyTree/DependencyTree.tsx#L186

src/components/DependencyTree/DependencyTree.tsx
```diff
@@ -0,0 +1,262 @@
...
+          level: level,
+        }));
+        if (currentNode.children) {
+          currentNode.children.concat(targetChildren);
+        } else {
+          currentNode.children = targetChildren;
+        }
+      } else {
+        edgesData!
+          .filter((edge) => (level === 1 && !isBuild(buildItem)) || edge.source === currentNode.id)
```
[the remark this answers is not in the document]
DnsZhou: > `(level === 1 && !isBuild(buildItem))` seems to be useless (if it was true, else branch would never execute)

The level can be other than 1. this is a recursive function so it will pass level+1 for every child of it. see https://github.com/DnsZhou/pnc-web-ui-react/blob/d3e71c56fc9b00c291355ff8c5e0ed2577543394/src/components/DependencyTree/DependencyTree.tsx#L186

src/components/DependencyTree/DependencyTree.tsx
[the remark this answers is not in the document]
patrikk0123: Yes, sure, but inside if, the level will always be equal to one, since it is condition required by if.

src/components/DependencyTree/DependencyTree.tsx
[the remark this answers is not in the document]
patrikk0123: What I was referring to is that if condition `(level === 1 && !isBuild(buildItem))` was true, `if branch` would execute and not `else branch`, in which this line of code is contained.

src/components/DependencyTree/DependencyTree.tsx
DnsZhou: So far there is no better solution since what we are doing here is to refresh the component using conditions. because patternfly do not provide a way to manipulate the expand status from outside after initialization.

src/components/DependencyTree/DependencyTree.tsx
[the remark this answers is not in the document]
DnsZhou: aha got it, missed that part

src/components/DependencyTree/DependencyTree.tsx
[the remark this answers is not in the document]
DnsZhou: okay got it

PR #88, opened 2022-08-18
[NCL-7175] Implement Log Viewer component

patrikk0123:


src/components/DemoPage/DemoPage.tsx
```diff
@@ -56,9 +60,47 @@ const formConfig = {
   },
 };
 
+const initLogData = [
```
DnsZhou, line 63: It would be better to create some timeout to display the demo log line by line, so that it would be more similar to a real live log on buildDetails page.

src/components/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    setLinesBehind(data.length - lineCount);
+  }, [data.length, lineCount]);
+
+  useEffect(() => {
+    if (isFollowing) {
+      setIsPaused(false);
+    }
+  }, [isFollowing, data]);
+
+  const onScroll = ({ scrollOffsetToBottom, scrollDirection, scrollUpdateWasRequested }: any) => {
```
DnsZhou, line 44: any type is not suggested, you can create an interface for that.

src/components/DemoPage/DemoPage.tsx
```diff
@@ -56,9 +60,47 @@ const formConfig = {
...
+  '[2022-08-15T14:11:37.018Z] Push ACCEPTED.',
+  '[2022-08-15T14:11:37.958Z] Applying tag fb-1.0-pnc on build org.jboss.modules-jboss-modules-1.5.0.Final_temporary_redhat_00033-1.',
+  '[2022-08-15T14:11:38.884Z] Sending callback to http://orch-master-devel.psi.redhat.com/pnc-rest/v2/builds/ATINBQK54TIAA/brew-push/complete.',
+  '[2022-08-15T14:11:38.91Z] Brew push completed.',
+];
+
 export const DemoPage = () => {
   useTitle('Demo Page');
 
+  const [logLine, setLogLine] = useState<any>('');
```
DnsZhou, line 101: any type is not suggested. try to avoid using that.

src/components/LogViewer/LogViewer.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { CaretDownIcon, CaretUpIcon, OutlinedPlayCircleIcon } from '@patternfly/react-icons';
+import { LogViewer as LogViewerPF } from '@patternfly/react-log-viewer';
+
+interface ILogViewerProps {
+  data: string | string[];
+  follow: boolean;
+}
+
+export const LogViewer = ({ data, follow }: ILogViewerProps) => {
+  const logViewerRef = useRef<any>();
```
DnsZhou, line 12: any type is not suggested

DnsZhou (changes requested):
Hi Patrik! 
Thank you for your work and here we have some improvements to be done. also, try to avoid using <any> type unless you have to. see detailed comments and let me know if you have any further questions.
-Dennis

PR #90, opened 2022-08-29
[NCL-7000] Implement PATCH method for Project update

patrikk0123:


jomrazek (commented):
Looks ok to me and works ok, we just talked with Patrik about returning back to project detail page from update page after update is done.

matedo1 (changes requested):

src/utils/patchHelper.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { IFieldValues, IFields } from '../containers/useForm';
+
+/**
+ * Transforms form state object into object of just form input values.
+ * Also filters all empty form inputs (empty values).
+ *
+ * @param data - form state
+ * @returns Object of non-empty form values
+ */
+export const createPatchData = (data: IFields): IFieldValues => {
```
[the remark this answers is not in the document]
patrikk0123: I decided to use `transformFormToValues` since patch is in this terminology array of operations in JSON patch format rather than object of values.

src/components/ProjectCreateEditPage/ProjectCreateEditPage.tsx
```diff
@@ -276,9 +302,16 @@ export const ProjectCreateEditPage = ({ editPage = false }: IProjectCreateEditPa
       <Flex direction={flexDirection}>
         <FlexItem>
           {editPage ? (
-            <DataContainer {...dataContainerCreateEdit} title="Edit Project">
-              {formComponent}
-            </DataContainer>
+            isPatching ? (
+              <ServiceContainerCreating {...dataContainerEditPatch} title="Edit Project">
```
matedo1, line 306: let's rename this from `ServiceContainerCreating` to `ServiceContainerCreatingUpdating`

src/services/projectService.ts
```diff
@@ -41,6 +42,18 @@ class ProjectService {
     return pncClient.getHttpClient().post('/projects', data, requestConfig);
   }
 
+  /**
+   * Patch a Project.
+   *
+   * @param id - project ID
+   * @param patchData - array of changes in JSON patch format
```
matedo1, line 49: Maybe we could keep naming consistent with the official documentation https://www.npmjs.com/package/fast-json-patch , so:
`... in JSON-Patch format`

matedo1 (changes requested):
just minor change, thanks!

matedo1 (approved):

PR #110, opened 2022-10-12
[NCL-7383] Create Editor Config file

patrikk0123:


DnsZhou (changes requested):
@patrikk0123 
 Hi Patrik,
Thanks for the work. According to the doc from editorConfig, it looks like we will need to install extra plugin to enable this editor config feature in VS Code(see https://editorconfig.org/). 

1. Ideally, we would like to have some editor config feature(for VS Code) that does not require any plugin, which means supported by VS Code naturally.
2. Or, if we can not find such a way, we will need to mention the installation of the plugin in the README doc of the repository, to add the plugin as one of the essential steps of setting up the development environment.

Feel free to reach me for any discussion if you want.
cc: @matedo1 

Thanks and regards,
- Dennis

patrikk0123:
> @patrikk0123 Hi Patrik, Thanks for the work. According to the doc from editorConfig, it looks like we will need to install extra plugin to enable this editor config feature in VS Code(see https://editorconfig.org/).
> 
> 1. Ideally, we would like to have some editor config feature(for VS Code) that does not require any plugin, which means supported by VS Code naturally.
> 2. Or, if we can not find such a way, we will need to mention the installation of the plugin in the README doc of the repository, to add the plugin as one of the essential steps of setting up the development environment.
> 
> Feel free to reach me for any discussion if you want. cc: @matedo1
> 
> Thanks and regards,
> 
> * Dennis

Hi Dennis,
I added documentation for this plugin. This plugin is pretty optional, so I think it should not matter that VS code plugin is needed to be installed.

matedo1:
What are actually minimal requirements to be able to fully use .editorconfig in VSCode? I can see for example https://stackoverflow.com/a/48523398 describing:

- extension
- npm package

are needed?

patrikk0123:
> What are actually minimal requirements to be able to fully use .editorconfig in VSCode? I can see for example https://stackoverflow.com/a/48523398 describing:
> 
> * extension
> * npm package
> 
> are needed?

just vs code plugin is needed

DnsZhou (approved):
lgtm, Thanks Patrik!

PR #116, opened 2022-11-01
[NCL-7388] Add ETA message component

DnsZhou:


src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -176,6 +178,40 @@ export const AdministrationPage = () => {
...
+                            const hour = date.getHours();
+                            const minute = date.getMinutes();
+                            const second = date.getSeconds();
+                            const monthString = month < 10 ? `0${month}` : month;
+                            const dayString = day < 10 ? `0${day}` : day;
+                            const hourString = hour < 10 ? `0${hour}` : hour;
+                            const minuteString = minute < 10 ? `0${minute}` : minute;
+                            const secondString = second < 10 ? `0${second}` : second;
+                            return `${year}-${monthString}-${dayString} ${hourString}:${minuteString}:${secondString}`;
+                          }}
```
patrikk0123, line 201: I would consider to create a separate method for this format inside utils/

matedo1 (dismissed):

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
+                      </GridItem>
+                    )}
+                    {isMaintenanceModeOn && (
+                      <GridItem span={12}>
+                        <FormGroup label="Example" fieldId="eta-time">
+                          {isMaintenanceModeOn && (
+                            <TopBarInfo>
+                              Maintenance Mode - PNC system is in the maintenance mode, no new build requests are accepted.
+                              Reason: {announcementMessage ? announcementMessage : 'N/A'}, ETA: {etaTime ? etaTime : 'N/A'}
+                            </TopBarInfo>
```
patrikk0123, line 279: There is a problem with this toolbar:
When you close the example toolbar and move to another page, for example Projects, also top bar on top side of page will disappear.
Solution:
Add parameter to the TopBar component which would make close button disabled (with default value enabled).

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
+                          )}
+                          {!isMaintenanceModeOn && <TopBarInfo>Announcement - {announcementMessage}</TopBarInfo>}
```
patrikk0123, line 281: Same as other top bar example.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
+                            label=" N/A:"
+                            labelOff=" N/A:"
+                            hasCheckIcon
+                            isChecked={isEtaNa}
+                            onChange={() => {
+                              setIsEtaNa(!isEtaNa);
+                              setEtaTouched(true);
+                            }}
+                            isReversed
+                          />
```
patrikk0123, line 268: Bug:
If date is already set and you switch N/A state to true, announcement will still include date and not 'N/A'.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
                               setIsMaintenanceModeOn(!isMaintenanceModeOn);
                             }}
                           />
                         </div>
                       </FormGroup>
                     </GridItem>
                     <GridItem span={12}>
-                      <FormGroup label="Announcement" fieldId="form-announcement">
+                      <FormGroup
+                        label={(isMaintenanceModeOn ? '* ' : '') + 'Announcement'}
```
patrikk0123, line 206: Use isRequired property on FormGroup insted of * character.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
+                          onBlur={() => {
+                            setAnnouncementTouched(true);
+                          }}
                         />
                       </FormGroup>
                     </GridItem>
+                    {isMaintenanceModeOn && (
+                      <GridItem span={12}>
+                        <FormGroup
+                          label="* ETA Time"
```
patrikk0123, line 228: Use isRequired property on FormGroup insted of * character.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -38,18 +43,48 @@ export const AdministrationPage = () => {
...
   useEffect(() => {
     genericSettingsService
       .getAnnouncementBanner()
       .then((response: any) => {
-        setAnnouncementMessage(response.data.banner);
+        const rawAnnouncement: string = response.data.banner;
+        const rawAnnouncementSet: Array<string> = rawAnnouncement ? rawAnnouncement.split(', ETA: ') : [];
+        setAnnouncementMessage(rawAnnouncementSet[0]);
+        if (rawAnnouncementSet[1]) {
+          if (rawAnnouncementSet[1] !== 'N/A') {
```
patrikk0123, line 79: Maybe these endpoints could be used for getting / setting maintanance mode:
/generic-setting/in-maintenance-mode
/generic-setting/activate-maintenance-mode
/generic-setting/deactivate-maintenance-mode
(maybe this belongs to other jira)

DnsZhou: This was a part of Implement top bar announcement messages [NCL-7284](https://issues.redhat.com/browse/NCL-7284) or Integrate top bar announcement messages with WebSockets [NCL-7317](https://issues.redhat.com/browse/NCL-7317) because maintenance mode is regarded as a kind of announcement message. and I discussed with martin that we should leave it to you to fix it.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -158,32 +193,107 @@ export const AdministrationPage = () => {
...
+                          helperTextInvalidIcon={<ExclamationCircleIcon />}
+                          validated={'error'}
+                        >
+                          <DatePicker
+                            required
+                            isDisabled={!isMaintenanceModeOn || isEtaNa}
+                            name="form-etaTime"
+                            id="eta-time"
+                            placeholder="yyyy-MM-dd hh:mm (UTC)"
+                            dateFormat={transDateFormat}
```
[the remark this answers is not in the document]
DnsZhou: I haven't found a better way to implement this since no format would look like YYYY-MM-DD HH:MM:SS

src/components/AdministrationPage/AdministrationPage.tsx
patrikk0123: Okay, I understand.

src/components/AdministrationPage/AdministrationPage.tsx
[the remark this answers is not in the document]
DnsZhou: YYYY-MM-DD HH:MM:SS might be the easiest way for the user to input, here the date format will cause when the user chooses any specific date with the date picker, it will automatically display the date as format defined here. If we instead ask the user to input a date with a format like `Dec 1, 2017, 2:17 PM`, there will be more room for mistakes by missing PM/AM or using blank space instead of `,`. I would suggest we actually use a format that is easiest to input and widely adopted by most of our users.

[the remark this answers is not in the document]
matedo1: yes, I agree, thanks for this point, let's keep it as it is

matedo1:
Hi @patrikk0123 , could you code review latest changes?

src/components/AdministrationPage/AdministrationPage.tsx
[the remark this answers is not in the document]
DnsZhou: fixed accordingly

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -38,18 +44,48 @@ export const AdministrationPage = () => {
...
+        if (rawAnnouncementSet[1]) {
+          if (rawAnnouncementSet[1] !== N_A) {
+            setEtaTime(rawAnnouncementSet[1]);
+            setIsEtaNa(false);
+          } else {
+            setEtaTime(undefined);
+            setIsEtaNa(true);
+          }
+        }
+        setIsMaintenanceModeOn(rawAnnouncementSet[1] ? true : false);
```
patrikk0123, line 88: Condition `rawAnnouncementSet[1] !== undefined` or even just `rawAnnouncementSet[1] ` would suffice I think.

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -38,18 +44,48 @@ export const AdministrationPage = () => {
...
+    return !isAnnouncementInvalid() && !isEtaTimeInvalid();
+  };
+
+  const isAnnouncementInvalid = () => {
+    return isMaintenanceModeOn && announcementTouched && (!announcementMessage || announcementMessage === '');
+  };
+
+  const isEtaTimeInvalid = () => {
+    return isMaintenanceModeOn && etaTouched && !isEtaNa && (!etaTime || etaTime === '');
+  };
```
patrikk0123, line 70: Have you considered using useForm hook?

patrikk0123 (changes requested):

src/components/AdministrationPage/AdministrationPage.tsx
DnsZhou: This is to include `rawAnnouncementSet[1] === null` as false as well.
Also rawAnnouncementSet[1] will not work:
image

src/components/AdministrationPage/AdministrationPage.tsx
DnsZhou: useForm is over-complicated because we are not using it in any fields of the administration page. we will also need to define formConfigs for all fields, at the same time we will need to refactor all filedId / id to lowerCamelCase naming because all fields are using kebab-case for now. since the admin page is for internal use only, If we decide to use it then we will also refactor it for PNC System Version fields. we can do it in a new Jira if really necessary.

src/components/AdministrationPage/AdministrationPage.tsx
matedo1: then consider using `!!rawAnnouncementSet[1]`

src/components/AdministrationPage/AdministrationPage.tsx
patrikk0123: fine

patrikk0123 (approved):
Now it looks fine, thanks.

src/components/AdministrationPage/AdministrationPage.tsx
DnsZhou: updated accordingly, Thanks Martin!

PR #137, opened 2022-12-07
[NCL-7459] Integrate WebSockets Client

patrikk0123:


patrikk0123:
The WS client has already been implemented now.

PR #147, opened 2023-01-23
[NCL-7538] Fix React state update on an unmounted component

patrikk0123:


src/hooks/useServiceContainer.ts
```diff
@@ -51,31 +60,35 @@ export const useServiceContainer = (service: Function, { initLoadingState = true
...
-              setError(error.toString());
-            }
+            setData(response.data);
+            setError(ERROR_INIT);
           });
+          return response;
+        }
+      })
+      .catch((error: any) => {
+        if (!isTerminated.current) {
```
matedo1, line 75: `if (error.name !== 'CanceledError') {` should be sufficient

src/hooks/useServiceContainer.ts
```diff
@@ -28,10 +28,19 @@ export const useServiceContainer = (service: Function, { initLoadingState = true
   const [data, setData] = useState<any>();
   const [loading, setLoading] = useState<boolean>(initLoadingState);
   const [error, setError] = useState<string>(ERROR_INIT);
+  // hook is terminated if unmounted from DOM
+  const isTerminated = useRef(false);
```
matedo1, line 32: I think this can be removed

matedo1 (changes requested):

matedo1 (approved):

PR #153, opened 2023-01-31
[NCL-7580] Update Axios to ^1.2.6

patrikk0123:


package.json
```diff
@@ -29,7 +29,7 @@
   "scripts": {
     "start": "REACT_APP_VERSION=$npm_package_version FAST_REFRESH=false react-scripts start",
     "build": "REACT_APP_VERSION=$npm_package_version react-scripts build",
-    "test": "react-scripts test",
+    "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!axios)/\"",
```
matedo1, line 32: Hi @DnsZhou , 

can you check this change it ok from your point of view? https://github.com/axios/axios/issues/5026#issuecomment-1271194848

DnsZhou: Hi @matedo1 ,as discussed in the axios issue,  after 1.0.0+ it refactored as an ES module, instead of common js module, which would cause issue with jest. I am fine with the workaround and I don't see any blocker as far as what we are working on, and I verified this solution with a clean build from my local env.

DnsZhou (approved):
Hi @patrikk0123 ,

I noticed the ESM conflict while running test, and I think the current solution is the best we can do, also we can see whether axios or Jest team would have some fix on that but before that we can keep with using `transformIgnorePatterns`.

arpitha-jaanaki:
Hello Team,
changing test script actually worked and fixed the issue like "import statement cannot be used outside the module"

but yet it is giving me other error like TypeError: Cannot set property 'Event' of undefined in event.js file 

event.js file looks like below

(function() {
  var extend = function(child, parent) { for (var key in parent) { if (hasProp.call(parent, key)) child[key] = parent[key]; } function ctor() { this.constructor = child; } ctor.prototype = parent.prototype; child.prototype = new ctor(); child.__super__ = parent.prototype; return child; },
    hasProp = {}.hasOwnProperty;

  this.Event = (function() {
    function Event(start_mark, end_mark) {
      this.start_mark = start_mark;
      this.end_mark = end_mark;
    }

    return Event;

  })();
  
 any idea on this,
 
 and i suspect axios is bringing this error. any help would be really appreciated
 
 thank you very much,

PR #159, opened 2023-02-08
[NCL-7581] Update chart.js to ^4.2.0

patrikk0123:


DnsZhou (approved):
Verified the Chart.js component with different set of data(mostly for build metrics), Thanks Patrik!

PR #160, opened 2023-02-13
[NCl-7591] Update prettier to 2.8.4

patrikk0123:


DnsZhou (approved):
Verified and works well, Thx Patrik!

PR #163, opened 2023-02-21
NCL-7392 Implement sortable table

matedo1:


patrikk0123 (approved):
Code seems fine and tested it locally. Approved.

PR #166, opened 2023-02-27
[NCL-7224] Implement Product Milestone detail page

patrikk0123:


src/components/ProductMilestoneDetailPage/ProductMilestoneDetailPage.tsx
```diff
@@ -1,9 +1,68 @@
...
+
+  const attributes = [
+    {
+      name: 'Status',
+      value: serviceContainerMilestone.data?.endDate ? 'CLOSED' : 'OPEN',
+    },
+    {
+      name: 'Start Date',
+      value: (
+        <OptionalText>{createDateTime({ date: serviceContainerMilestone.data.startingDate, includeTime: false })}</OptionalText>
```
DnsZhou, line 40: The date format is different from our angularjs version, are we going to refactor them in a new way or we should keep it as the same one with before?
image
image

@matedo1

patrikk0123: This format is used in **createDateTime** function so I think it is correct.

DnsZhou (approved):
All the code part looks good to me except the date format.

matedo1 (changes requested):
Just small changes, thanks

matedo1 (approved):

PR #168, opened 2023-02-28
[NCL-7629] Unable to go back in browser history on Projects page

patrikk0123:


matedo1 (approved):
Thanks for noticing and fixing it!

PR #174, opened 2023-03-14
[NCL-7622] Make AttributesItems declarative

patrikk0123:


src/components/AttributesItems/DeclarativeAttributesItems.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const AttributesItem = ({ children, title, tooltip }: PropsWithChildren<IAttributesItemProps>) => (
+  <>
+    <GridItem xl={3} lg={4} md={6} sm={12} className={styles['name']}>
+      <>
+        {title}
+        {tooltip && <TooltipWrapper tooltip={tooltip} />}
+      </>
+    </GridItem>
+    <GridItem xl={9} lg={8} md={6} sm={12} className={css(!children && styles['value-empty'])}>
+      {children ? children : 'Empty'}
```
matedo1, line 36: `EmptyStateSymbol` will be used instead of "Empty"

src/components/AttributesItems/DeclarativeAttributesItems.tsx
```diff
@@ -0,0 +1,40 @@
...
+import styles from './AttributesItems.module.css';
+
+interface IAttributesItemsProps {
+  isGridUniform?: boolean;
+}
+
+/**
+ * Represents a stylized name-value table component.
+ */
+export const AttributesItems = ({ children, isGridUniform = true }: PropsWithChildren<IAttributesItemsProps>) => (
```
DnsZhou, line 17: Personally I would suggest using `Attributes` and `AttributesItem`, so that it would be similar with PatternFly naming styles(e.g., `Flex`/`FlexItem`, `ToolBar`/`ToolBarItem`. This would be clearer than `AttributesItems`/`AttributesItem`, which looks too similar.

DnsZhou (changes requested):
Hi @patrikk0123 , just a minor suggestion on the naming of the component. Otherwise it LGTM.

btw @matedo1 , are we going to refactor to use DeclarativeAttributesItems instead of the original integration(passing const attributes) in the future?

src/components/AttributesItems/DeclarativeAttributesItems.tsx
matedo1: naming - yes, you are right and we already discussed it with patrik (see my comments above), it's just temporary solution

src/components/AttributesItems/DeclarativeAttributesItems.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const AttributesItem = ({ children, title, tooltip }: PropsWithChildren<IAttributesItemProps>) => (
+  <>
+    <GridItem xl={3} lg={4} md={6} sm={12} className={styles['name']}>
+      <>
+        {title}
+        {tooltip && <TooltipWrapper tooltip={tooltip} />}
+      </>
+    </GridItem>
+    <GridItem xl={9} lg={8} md={6} sm={12}>
+      {children ? children : <EmptyStateSymbol variant="text" />}
```
matedo1, line 37: Did you test it works successfully with our common scenarios?

matedo1 (changes requested):
Just small changes, thanks, let's do one more code-review and after that you can integrate it everywhere.

src/components/AttributesItems/DeclarativeAttributesItems.tsx
patrikk0123: Yes I tested it with undefined, empty string, etc.

matedo1 (approved):

PR #176, opened 2023-03-14
[NCL-7667] Artifact list row expanding

patrikk0123:


DnsZhou (commented):
Hi @patrikk0123 ,

Thanks for the work. looks like now we are passing `onFilter` function to the filtering component, and the specific case for this scenario is manually trigger `setAreAllArtifactsExpanded(true)` so that all results will be expanded. Do you think we need to trigger collapse all when there are no target filters (md5,sha1,sha256)?

Everything else looks good to me other than that. Thanks!

patrikk0123:
> Hi @patrikk0123 ,
> 
> Thanks for the work. looks like now we are passing `onFilter` function to the filtering component, and the specific case for this scenario is manually trigger `setAreAllArtifactsExpanded(true)` so that all results will be expanded. Do you think we need to trigger collapse all when there are no target filters (md5,sha1,sha256)?
> 
> Everything else looks good to me other than that. Thanks!

I don't think they should collapse when you are filtering by something other than hash. You can still press 'Collapse All' button.

matedo1:
Is `areAllArtifactsExpanded` + `useEffect` logic really needed? I think mostly you could do do same job without it and I think it could also reflect the current state more precisely:
 - `expandedArtifacts.length === serviceContainerArtifacts.data?.content.length` // everything is expanded
 - `expandedArtifacts.length === 0` // nothing is expanded
 - etc.

Read also https://beta.reactjs.org/learn/you-might-not-need-an-effect , especially read 
image

patrikk0123:
> > Hi @patrikk0123 ,
> > Thanks for the work. looks like now we are passing `onFilter` function to the filtering component, and the specific case for this scenario is manually trigger `setAreAllArtifactsExpanded(true)` so that all results will be expanded. Do you think we need to trigger collapse all when there are no target filters (md5,sha1,sha256)?
> > Everything else looks good to me other than that. Thanks!
> 
> I don't think they should collapse when you are filtering by something other than hash. You can still press 'Collapse All' button.



> Is `areAllArtifactsExpanded` + `useEffect` logic really needed? I think mostly you could do do same job without it and I think it could also reflect the current state more precisely:
> 
> * `expandedArtifacts.length === serviceContainerArtifacts.data?.content.length` // everything is expanded
> * `expandedArtifacts.length === 0` // nothing is expanded
> * etc.
> 

I tried your solution, but mine is better since it keeps all rows expanded even after chaning list page.



> Is `areAllArtifactsExpanded` + `useEffect` logic really needed? I think mostly you could do do same job without it and I think it could also reflect the current state more precisely:
> 
> * `expandedArtifacts.length === serviceContainerArtifacts.data?.content.length` // everything is expanded
> * `expandedArtifacts.length === 0` // nothing is expanded
> * etc.


I tried your solution, but mine is better since it keeps all rows expanded even after chaning list page.

patrikk0123:
> Is `areAllArtifactsExpanded` + `useEffect` logic really needed? I think mostly you could do do same job without it and I think it could also reflect the current state more precisely:
> 
> * `expandedArtifacts.length === serviceContainerArtifacts.data?.content.length` // everything is expanded
> * `expandedArtifacts.length === 0` // nothing is expanded
> * etc.

I tried your solution, but mine is better since it keeps all rows expanded even after chaning list page.

matedo1 (approved):

DnsZhou (approved):

DnsZhou:
lgtm, Thanks @patrikk0123 !

PR #178, opened 2023-03-14
[NCL-7224] Implement Milestone Delivered Artifacts page

patrikk0123:


DnsZhou (changes requested):
There seems to be a sort error, when sort was triggered, the page returns no data. see screen shots:
patrikk0123:ncl7224-deliveredArtifactsPage
other than that, the code looks good.
Thanks!

patrikk0123:
> There seems to be a sort error, when sort was triggered, the page returns no data.

Thanks for pointing it out, I rebased branch to use new Artifact Table version which does not have Build column, so this should no longer be a problem.

DnsZhou (approved):

src/services/productMilestoneApi.tsx
```diff
@@ -36,14 +36,23 @@ export const getProductMilestoneDeliveredArtifacts = (
...
  *
  * @param requestConfig - Axios based request config
  */
 export const getCloseResults = ({ id }: IProductMilestoneApiData, requestConfig: AxiosRequestConfig = {}) => {
   return pncClient.getHttpClient().get<ProductMilestoneCloseResultPage>(`/product-milestones/${id}/close-results`, requestConfig);
 };
 
+/**
+ * Gets Product Milestone Delivered Artifacts.
+ *
```
matedo1, line 49: Add service data param

PR #202, opened 2023-04-19
[NCL-7638] Add cancel sorting

DnsZhou:


matedo1 (changes requested):
You commit message and PR name are wrong, you wanted to use `Add cancel sorting` instead.

You are going in a good direction, but some additional changes and updates needs to be applied.

DnsZhou:
Will change the PR name and jira name as well. We had a wrong jira name as well...

DnsZhou:
Choosed solution 1 and renamed the branch, thus created another PR instead: https://github.com/project-ncl/pnc-web-ui-react/pull/205

PR #209, opened 2023-04-24
[NCL-7210] Implement Build detail pages structure 

patrikk0123:


matedo1 (approved):
Entity name related components / functions like

- calculateBuildName
- calculateBuildConfigName
- etc

will be refactored (see NCL-7814) later to reflects our latest decisions. Keep those changes minimal as much as possible.

PR #210, opened 2023-04-25
[NCL-7799] Implement Build Artifact Dependency Graph

patrikk0123:


src/hooks/useNetworkGraph.ts
```diff
@@ -254,7 +254,8 @@ export const useNetworkGraph = ({
     renderer.current?.setSetting('nodeReducer', (node: string, data: any) => {
       const nodeAttribute = graph.current?.getNodeAttributes(node);
 
-      const labelMatches = searchValueMainLabel && node.toLowerCase().includes(searchValueMainLabel.toLowerCase());
+      const labelMatches =
```
DnsZhou, line 257: Would suggest `isLabelMatch` instead of `labelMatches` as variable name to meke it clear that this is a boolean value. Using `is` as a prefix is widely used in our repo as well.

src/hooks/useNetworkGraph.ts
```diff
@@ -254,7 +254,8 @@ export const useNetworkGraph = ({
...
+        searchValueMainLabel && nodeAttribute?.mainLabel?.toLowerCase().includes(searchValueMainLabel.toLowerCase());
       const subLabelMatches =
```
DnsZhou, line 259: see above, suggest `isSubLabelMatch`

src/hooks/useNetworkGraph.ts
```diff
@@ -254,7 +254,8 @@ export const useNetworkGraph = ({
...
         searchValueSubLabel && nodeAttribute?.subLabel?.toLowerCase().includes(searchValueSubLabel.toLowerCase());
       if (
```
DnsZhou, line 261: see above, suggest `isLabelMatch` and `isSubLabelMatch`

src/components/ServiceContainers/ServiceContainerLoading.tsx
```diff
@@ -68,7 +68,7 @@ export const ServiceContainerLoading = ({
   // (the component was rendered at some point before)
   //  - for example: when page index is changed from page 1 to page 2
   //  - this will make UI more smooth and it prevents flickering user experience
-  if (loading && data) return <RefreshStateCard>{children}</RefreshStateCard>;
+  if (loading && data) return <RefreshStateCard isInline={variant !== 'block'}>{children}</RefreshStateCard>;
```
DnsZhou, line 71: Put some comments here to describe the reason for this bug fix

DnsZhou (changes requested):
Looks good to me and verified locally. Just some naming suggestions. Thanks @patrikk0123

src/components/StateCard/RefreshStateCard.tsx
```diff
@@ -1,13 +1,20 @@
...
  */
-export const RefreshStateCard = ({ children }: React.PropsWithChildren<IRefreshStateCard>) => (
-  <div className={styles['refresh-state-card']}>
-    <div className={styles['refresh-state-card__body']}>{children}</div>
-    <div className={styles['refresh-state-card__indicator']}></div>
-  </div>
-);
+export const RefreshStateCard = ({ children, isInline = false }: React.PropsWithChildren<IRefreshStateCard>) =>
+  isInline ? (
+    <Spinner isInline isSVG aria-label="Loading..." />
```
matedo1, line 14: There are already 3 places where Spinner is used (search `<Spinner`).

Let's create `component/Spinner/Spinner.tsx` and reuse it everywhere. Import PatternFly  spinner internally as `SpinnerPF`.

matedo1 (changes requested):
(Note: this implementation is reusing the logic from #190 , see that for more details)

Thanks!

src/hooks/useNetworkGraph.ts
patrikk0123: Okay, but I used `doesLabelMatch`.

src/components/ServiceContainers/ServiceContainerLoading.tsx
patrikk0123: When `inline` variant of `ServiceContainerLoading` is used and data of the service container are reloaded, `block` variant of the loading bar was displayed. Instead, a small inline spinner needs to be shown.

matedo1 (approved):
@DnsZhou WDYT about `doesLabelMatch` ?

DnsZhou:
I would still suggest using 'is' because that more applies to best practices. See online discussion  here: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

patrikk0123:
> 

It' just one of the opinions. `isLabelMatch` does not sound right grammatically.

DnsZhou (approved):

PR #214, opened 2023-05-02
[NCL-7688] Add Product Version description on Product detail page

DnsZhou:


matedo1 (changes requested):
I put a comment into Jira.

src/components/ProductDetailPage/ProductDetailPage.tsx
```diff
@@ -76,9 +76,13 @@ export const ProductDetailPage = ({ componentId = 'v1' }: IProductDetailPageProp
         </ContentBox>
 
         <Toolbar>
-          <ToolbarItem>
+          <ToolbarItem reservedWidth>
```
matedo1, line 79: How this would be implemented if more `ToolbarItem` is used (for example see `ArtifactsList.tsx`)?

patrikk0123 (approved):

src/components/ProductDetailPage/ProductDetailPage.tsx
DnsZhou: @matedo1  it also works well, if you want to reserve width to allow suffix components/buttons not to be wrapped, just add `reserveWidth` to the current component. see pictures:
image
image

DnsZhou: image

matedo1 (approved):

PR #222, opened 2023-05-11
[NCL-7224] Implement Builds Performed page

patrikk0123:


DnsZhou (changes requested):
Just some changes over the JSDoc, it looks good other than that. 

Also, There is an error while directing to the second page of the builds performed page, this seems to be a backend issue. You can create a Jira if there isn't one for this issue. But nothing need to be done since this is not a front-end issue.
(e.g. for the bug: http://localhost:3000/pnc-web/products/100/versions/100/milestones/101/builds-performed?b1-sort=%3Ddesc%3DsubmitTime&b1-pageIndex=2&b1-pageSize=10)

matedo1 (approved):

patrikk0123:
Backend issue was now already fixed AFAIK.

PR #236, opened 2023-06-19
[NCL-7923] Kill the graph resources in the cleanup function

patrikk0123:


DnsZhou (changes requested):
Hi @patrikk0123 , seems the issue remains there. See jira NCL-7923 for screen recordings.

PR #244, opened 2023-06-29
[NCL-7980] Integrate entity attributes into Product pages

patrikk0123:


patrikk0123:
@matedo1 Are tooltips needed for titles of these attributes?

matedo1:
@patrikk0123 Let's skip them for now, there is just one rule - if tooltips are available on one page (for example create page), then they have to applied also on all other pages (detail, list, etc).

matedo1 (changes requested):
Thanks! I was able to find some additional places:
- ProductMilestoneReleaseLabel.tsx

matedo1 (approved):
Thanks!

PR #248, opened 2023-07-04
[NCL-7666] Fix filtering layout

patrikk0123:


src/components/ArtifactsList/ArtifactsList.tsx
```diff
@@ -163,7 +163,10 @@ export const ArtifactsList = ({ serviceContainerArtifacts, columns = defaultColu
             }}
           />
         </ToolbarItem>
-        <ToolbarItem marginLeft="20px">
+      </Toolbar>
+
+      <Toolbar padding={false} paddingBottom paddingLeft paddingRight>
```
matedo1, line 168: Maybe the opposite could be considered:
 - apply padding automatically 
 - disable one (both) of them manually

My idea is that:
1) syntax is shorter
2) I assume it will be more typical for other use-cases too

```tsx
<Toolbar disablePaddingTop>

// next use cases (at this moment I think there should be no need to modify left + right):
<Toolbar disablePaddingTop disablePaddingBottom>
```

Let me know if it's problematic for our current codebase to implement it like this.

patrikk0123: It's shorter, but I found the semantic of removing the padding instead of adding it as it's needed, weird.
Maybe I could name those properties in shorthand form : `p` for padding, `pl` for padding left.

matedo1: I would still go with full naming in this case, it means `paddingLeft` instead of `pl`, the thing is that left + right should be applied always, or is there any case where it's not needed?

patrikk0123: Okay, the long format of names might stay, but I would not go with removing instead of adding padding.
Right now, it is shorter to just remove it on top/bottom, but generally, it does not look right. In other situations, you might want a different padding combination.

On the Toolbar component, it would be quite acceptable to use your variant, since it has less variable use cases, but ContentBox is general.

matedo1: Yes, if no other issues occur, let's try following combination:
1) `<Toolbar disablePaddingTop disablePaddingBottom>` // specialized on current usecases
2) `<ContentBox paddingTop paddingBottom paddingLeft paddingRight>` // generalized

If there is some other issue (for example you would feel it too inconsistent), let's implement your original solution.

patrikk0123: I implemented this combined solution.

matedo1 (approved):

PR #249, opened 2023-07-05
[NCL-7898] Refactor useForm to add boolean value for preBuildSyncEnable

DnsZhou:


DnsZhou:
Known issue: there is an uncaught type error with patch requests in useForm hook, e.g., edit scmRepository scenario or edit projects. But the patch request itself finished properly.

patrikk0123 (approved):

PR #250, opened 2023-07-07
[NCL-7759] Make Tabs component responsive

DnsZhou:


src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,45 @@ interface ITabsProps {
...
+      tabContent.scrollLeft = tabContent.scrollLeft ? tabContent.scrollLeft - 150 : 0;
+    }
+  };
+  const scrollRight = () => {
+    if (tabContent) {
+      tabContent.scrollLeft = tabContent.scrollLeft ? tabContent.scrollLeft + 150 : 150;
+    }
+  };
+
+  useEffect(onResize, [onResize]);
```
matedo1, line 36: I have several points:
1) you could avoid `onResize` in dependency array
2) you miss remove event listener
3) `getElementById` should be used after component is rendered, also check if `useRef` should be used instead?

I am providing some draft for you:
```tsx
useEffect(() => {
  const onResize = /* ... */ 

  tabContent = /* ... */
  window.addEventListener('resize', onResize);
  return () => {
    window.removeEventListener('resize', onResize);
  }
}, []);
```

matedo1 (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,45 @@ interface ITabsProps {
...
+
+  const onResize = useCallback(() => {
+    tabContent && setShowScrollButton(tabContent.scrollWidth > window.innerWidth);
+  }, [tabContent]);
+
+  window.addEventListener('resize', onResize);
+
+  const scrollLeft = () => {
+    if (tabContent) {
+      tabContent.scrollLeft = tabContent.scrollLeft ? tabContent.scrollLeft - 150 : 0;
```
patrikk0123, line 27: If the the `scrollLeft` is zero, 'scroll left' button should be disabled.
Same for the 'scroll right' variant.
In the same way as here:
https://www.patternfly.org/v4/components/tabs/

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,45 @@ interface ITabsProps {
...
+    }
+  };
+  const scrollRight = () => {
```
patrikk0123, line 30: Missing empty line.

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,45 @@ interface ITabsProps {
...
 
 export const Tabs = ({ children }: ITabsProps) => {
+  const [showScrollButton, setShowScrollButton] = useState(false);
+  const tabContent = document.getElementById('tab-content');
+
+  const onResize = useCallback(() => {
+    tabContent && setShowScrollButton(tabContent.scrollWidth > window.innerWidth);
+  }, [tabContent]);
+
+  window.addEventListener('resize', onResize);
```
patrikk0123, line 23: This should be executed inside an useEffect lest the event handler is not assigned on each render.
```
useEffect(() => {
  window.addEventListener('resize', onResize);
}, [onResize]);
````

Although,  `const tabContent = document.getElementById('tab-content');` needs to be moved inside `onResize` callback.

File: src/components/Tabs/Tabs.tsx [hunk 5]
```diff
@@ -12,15 +13,45 @@ interface ITabsProps {
 }
 
 export const Tabs = ({ children }: ITabsProps) => {
+  const [showScrollButton, setShowScrollButton] = useState(false);
+  const tabContent = document.getElementById('tab-content');
+
+  const onResize = useCallback(() => {
+    tabContent && setShowScrollButton(tabContent.scrollWidth > window.innerWidth);
+  }, [tabContent]);
```
patrikk0123, line 21: `tabContent` is assigned on each render, so onResize is recreated on each render.
Then useEffect handler below is rerun on each render as well.
Isn't `resize` event listener enough?

patrikk0123 (commented):
I tested it and Tabs are responsive, but the buttons should be disabled when the scroll is in its left/right-most position.

src/components/Tabs/Tabs.tsx
DnsZhou: the reason for calling `onResize` here is to initialize the scroll button when first loaded. Narrow-screen user will not be able to see the button without calling it at the very beginning(because they actually never "resize" the window.

DnsZhou:
see jira NCL-7759 for latest gifs

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,68 @@ interface ITabsProps {
 }
 
 export const Tabs = ({ children }: ITabsProps) => {
+  const [showScrollButton, setShowScrollButton] = useState(false);
+  const [disableScrollLeft, setDisableScrollLeft] = useState(true);
+  const [disableScrollRight, setDisableScrollRight] = useState(false);
+  const tabContent = useRef<HTMLUListElement>(null);
+
+  const onResize = useCallback(() => {
```
matedo1, line 21: Why not to put `onResize` to `useEffect` directly to avoid additional dependencies, `useCallback` usage, etc?

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,68 @@ interface ITabsProps {
 }
 
 export const Tabs = ({ children }: ITabsProps) => {
+  const [showScrollButton, setShowScrollButton] = useState(false);
+  const [disableScrollLeft, setDisableScrollLeft] = useState(true);
+  const [disableScrollRight, setDisableScrollRight] = useState(false);
```
matedo1, line 18: We agreed to use slightly different naming conventions for cases like this, `isScrollRightDisabled`, etc.

Also we use types `useState<boolean>`.

At same point we can revalidate our decision again, but for now we should be consistent.

src/components/Tabs/Tabs.tsx
```diff
@@ -12,15 +13,68 @@ interface ITabsProps {
...
+
+  const onResize = useCallback(() => {
+    tabContent.current && setShowScrollButton(tabContent.current.scrollWidth > window.innerWidth);
+  }, [tabContent]);
+
+  const scrollLeft = () => {
+    if (tabContent.current) {
+      setDisableScrollRight(false);
+      setDisableScrollLeft(tabContent.current.scrollLeft - 150 <= 0);
+      tabContent.current.scrollLeft = tabContent.current.scrollLeft ? tabContent.current.scrollLeft - 150 : 0;
```
matedo1, line 29: `150` should be replaced by some constant, for example `SCROLL_SIZE`, then it should be reused everywhere

matedo1 (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -14,26 +14,37 @@ interface ITabsProps {
...
     }
   };
+
   const scrollRight = () => {
-    if (tabContent) {
-      tabContent.scrollLeft = tabContent.scrollLeft ? tabContent.scrollLeft + 150 : 150;
+    if (tabContent.current) {
+      setDisableScrollLeft(false);
+      setDisableScrollRight(tabContent.current.scrollWidth <= window.innerWidth + tabContent.current.scrollLeft + 150);
```
patrikk0123, line 36: `scrollLeft` / `scrollRight` callbacks are run only when respective buttons are pressed.
This does not take into mind the usage of the touchpad / mouse scroll button.
Therefore, these two set functions need to be moved to some scroll event listener; solution below should work.

```
<ul
  ref={tabContent}
  onScroll={() => {
    if (tabContent.current) {
      setDisableScrollLeft(tabContent.current.scrollLeft - 5 <= 0);
      setDisableScrollRight(
        tabContent.current.scrollLeft + tabContent.current.offsetWidth + 5 >= tabContent.current.scrollWidth
      );
    }
  }}
  className={stylesPF.tabsList}
>
  {children}
</ul>
```

+/- 5 just due to he inability to scroll to the far right/left on certain screen sizes.

DnsZhou: Got it, thanks for raising up the solution!

src/components/Tabs/Tabs.tsx
```diff
@@ -1,26 +1,98 @@
...
 
+const SCROLL_BUTTON_SIZE = 150;
+const SCROLL_TOUCH_SIZE = 5;
+
 interface ITabsProps {
   children: ReactElement | ReactElement[];
 }
 
 export const Tabs = ({ children }: ITabsProps) => {
+  const [isScrollButtonDisplay, setIsScrollButtonDisplay] = useState<boolean>(false);
```
patrikk0123, line 19: `isScrollButtonDisplayed` / `isScrollButtonShown`

src/components/Tabs/Tabs.tsx
```diff
@@ -1,26 +1,98 @@
...
 import { css } from '@patternfly/react-styles';
 import stylesPF from '@patternfly/react-styles/css/components/Tabs/tabs';
-import { ReactElement } from 'react';
+import { ReactElement, useEffect, useRef, useState } from 'react';
 
 import styles from './Tabs.module.css';
 
 const stickyOnBreakpoint: PageSectionProps['stickyOnBreakpoint'] = { default: 'top' };
 
+const SCROLL_BUTTON_SIZE = 150;
```
patrikk0123, line 11: `SCROLL_OFFSET`

DnsZhou:
Hi @patrikk0123 , feel free to verify the changes. Thanks.

patrikk0123 (approved):
I checked the code and verified the solution locally, it now works correctly.
Thanks, but please squash the commits into just one.

PR #255, opened 2023-07-17
[NCL-7929] Replace onKeyPress with onKeyDown

DnsZhou:


patrikk0123 (approved):
`keypress` event is deprecated now, indeed, so the change seems reasonable to me.
https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event

PR #256, opened 2023-07-18
NCL-8009 Implement service container data values

matedo1:


DnsZhou (approved):
Hi @matedo1 , I tested the changes on my local environment with related pages such as Product Milestone Comparation insights and CreatEdit pages, all work good without any errors. Also, I checked all the changed parts and understand the changes, will also apply the same to new pages in the future. Thanks.

PR #257, opened 2023-07-19
[NCL-7659] Set the width of ContentBox responsively

DnsZhou:


matedo1:
During the demo meeting we agreed on additional changes to be applied

matedo1 (changes requested):
I think there are several other pages where `isResponsive` needs to be applied, right?

For example all Product related detail pages, Project detail page, Admin page, About page, etc. 

Please make sure to go through all pages and test them.

I would skip Variables page.

DnsZhou:
Hi @matedo1 , following pages have also been integrated with the responsive feature:

AboutPage
AdministrationPage
ProductDetailPage
ProductMilestoneDetailPage (just for Attributes)
ProductVersionDetailPage (just for Attributes)
ProjectDetailPage

src/components/ContentBox/ContentBox.tsx
```diff
@@ -18,6 +19,7 @@ interface IContentBoxProps {
   title?: string;
   contentHeight?: string;
   description?: IDescription;
+  isResponsive?: boolean;
```
matedo1, line 22: Add some TSDoc explaining when it should be used.

matedo1 (changes requested):

src/components/ContentBox/ContentBox.tsx
DnsZhou: Added accordingly

matedo1 (approved):

PR #259, opened 2023-07-22
[NCL-7613] Fix act warnings for unit test

DnsZhou:


patrikk0123 (approved):
I reviewed the changes and they seem fine to me, even though my knowledge of Jest is limited.
I tested it locally, and only the `ErrorBoundary` console errors remain, which is to be expected.

PR #261, opened 2023-07-26
[NCL-7204] BuildConfigList page initial commit without BuildStatus

DnsZhou:


src/common/buildConfigEntityAttributes.ts
```diff
@@ -0,0 +1,52 @@
+import { BuildConfiguration } from 'pnc-api-types-ts';
+
+import { TEntityAttributes } from 'common/entityAttributes';
+
+const buildTypeValues: BuildConfiguration['buildType'][] = ['MVN', 'NPM', 'GRADLE', 'SBT'];
+
+interface IExtendedBuildConfig extends BuildConfiguration {
+  buildStatus: any;
```
matedo1, line 8: Reuse `buildStatus` type: https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/common/buildEntityAttributes.ts#L5

DnsZhou:
Clean build has failed, it is most likely caused by upstream projects. Checking in progress...

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -0,0 +1,249 @@
...
+                                  </span>
+                                )}
+                              </DescriptionListDescription>
+                            </DescriptionListGroup>
+                          )}
+                          {columns.includes(buildConfigEntityAttributes.modificationTime.id) && (
+                            <DescriptionListGroup>
+                              <DescriptionListTerm>{buildConfigEntityAttributes.modificationTime.title}</DescriptionListTerm>
+                              <DescriptionListDescription>
+                                {buildConfig.modificationTime && createDateTime({ date: buildConfig.modificationTime }).custom}
```
patrikk0123, line 207: try to use shorter date format in certain conditions with the `includeDateInCustom` using the same logic as that of BuildsList times column

src/common/buildConfigEntityAttributes.ts
```diff
@@ -0,0 +1,65 @@
...
+    title: 'Modified',
+  },
+  creationUser: {
+    id: 'creationUser',
+    title: 'Created By',
+  },
+  modificationUser: {
+    id: 'modificationUser',
+    title: 'Modified By',
+  },
```
patrikk0123, line 56: these could also be filterable

src/common/buildConfigEntityAttributes.ts
```diff
@@ -0,0 +1,65 @@
...
+    title: 'Build Type',
+    values: buildTypeValues,
+    filter: {
+      operator: '==',
+    },
+  },
+  project: {
+    id: 'project',
+    title: 'Project',
+  },
```
patrikk0123, line 40: this might be filterable i think

src/common/buildConfigEntityAttributes.ts
DnsZhou: See Jira comments, these are not supported

patrikk0123 (approved):

src/index.css
```diff
@@ -153,6 +153,13 @@ code {
   background-color: #e9e9e9 !important;
 }
 
+/* 
+ * table alignment behaviour
+ */
+.pf-c-table tbody > tr > * {
```
matedo1, line 159: Let's remove it for now, but we can revalidate it again in the future

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -0,0 +1,254 @@
...
+                        columns.includes(buildConfigEntityAttributes.creationTime.id)
+                          ? sortAttributes.creationTime.id!
+                          : sortAttributes.modificationTime.id!
+                      )}
+                      isDropdownOpen={isTimesSortDropdownOpen}
+                      onDropdownToggle={() => setIsTimesSortDropdownOpen(!isTimesSortDropdownOpen)}
+                    />
+                  </Th>
+                )}
+                {columns.includes(buildConfigEntityAttributes.buildStatus.id) && (
```
matedo1, line 154: After the discussion we could see that Latest Build column introduces a lot of other issues, so we decided to remove it for now. It will be implemented in new separated jira. See jira comments.

matedo1 (changes requested):

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  <Th width={25}>{buildConfigEntityAttributes.buildStatus.title}</Th>
+                )}
+                {columns.includes(buildConfigEntityAttributes.actions.id) && (
+                  <ProtectedComponent hide>
```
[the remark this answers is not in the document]
matedo1: @DnsZhou can you please create new for ProtectedContent?

src/common/buildConfigEntityAttributes.ts
```diff
@@ -0,0 +1,68 @@
...
+  modificationTime: {
+    id: 'modificationTime',
+    title: 'Modified',
+  },
+  creationUser: {
+    id: 'creationUser',
+    title: 'Created By',
+  },
+  modificationUser: {
```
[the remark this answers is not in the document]
patrikk0123: I would still keep it there, for future or possible purposes.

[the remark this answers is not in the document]
DnsZhou: we can add it back in the future if needed, but for now, we will remove it to keep the code neat.

src/components/BuildConfigsList/BuildConfigsList.tsx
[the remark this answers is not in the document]
DnsZhou: NCL-8095 created for this issue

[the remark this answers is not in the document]
DnsZhou: The action won't pass to the backend anyway when not logged in, because the backend will return a 401 error. we can keep it like this to automatically fix it when NCL-8095 finishes.

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    {buildConfigEntityAttributes['project.name'].title}
+                  </Th>
+                )}
+                {filteredColumns.includes(buildConfigEntityAttributes.creationTime.id) &&
+                  filteredColumns.includes(buildConfigEntityAttributes.modificationTime.id) && (
+                    <Th width={25} className="overflow-visible">
+                      <SortGroup
+                        title="Times"
+                        sort={getSortGroupParams(
+                          filteredColumns.includes(buildConfigEntityAttributes.creationTime.id)
```
matedo1, line 150: This condition is not needed, use just `sortAttributes.creationTime.id`

matedo1 (changes requested):

src/utils/utils.ts
```diff
@@ -148,3 +150,23 @@ export const calculateFileSize = (sizeInBytes: number): string => {
...
+ * Function to check if the required combination has been met.
+ *
+ * @param columns - columns to be checked
+ * @param combinations - the required combination that need to exist together
+ *
+ */
+export const checkColumnsCombinations = ({ columns, combinations }: ICheckColumnsCombinations) => {
+  let isMismatch = false;
+  combinations.forEach((combination) => {
+    isMismatch = isMismatch || (combination.length >= 2 && combination.some((element) => !columns.includes(element)));
```
matedo1, line 169: You are going through all combinations even if you know for example the first one already failed.

src/components/TruncatedUsername/TruncatedUsername.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * The username component that will display a truncated username.
+ *
+ * @param text - The raw text of the username to be truncated
+ * @param length - The target length for usernames that are too long, by default 20
+ */
+export const TruncatedUsername = ({ text, length = 20 }: ITruncatedTextProps) => {
+  const removeKeyword = 'service-account-';
+  let resultText = text.replace(removeKeyword, '');
```
matedo1, line 18: I noticed 2 things:
1) much simpler syntax can be used (for `substring` - any argument value that is less than 0 or greater than str.length is treated as if it were 0 and str.length, respectively)
2) `length - 1` is wrong, only `length` should be used, otherwise it's not target length (see your TSDoc)

```tsx
const resultText = text.replace(removeKeyword, '').substring(0, length);

return ...
```

matedo1 (changes requested):

src/components/TruncatedText/TruncatedText.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * Truncate text component that will truncate the children as text and  that redirect to specific page.
+ *
+ * @param parsedUrl - the object contains url and name
+ * @param isInline - whether to use inline style with external link action
+ */
+export const TruncatedText = ({ length = 20, children }: PropsWithChildren<ITruncatedTextProps>) => {
+  const text = typeof children === 'string' ? (children as string) : null;
+  return text && text.length >= length ? (
+    <TooltipWrapper tooltip={text}>
+      <span>{text.substring(0, length - 1)} ...</span>
```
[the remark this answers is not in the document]
DnsZhou: `<span>` is necessary because otherwise, it can not trigger the tooltip when hovering on the truncated username. `<></>` does not work either.

src/utils/utils.ts
```diff
@@ -148,3 +150,27 @@ export const calculateFileSize = (sizeInBytes: number): string => {
...
+
+/**
+ * Function to check if the combination requirements has been satisfied.
+ *
+ * @param columns - columns to be checked
+ * @param combinations - the combination requirements that need to be satisfied
+ *
+ */
+export const checkColumnsCombinations = ({ columns, combinations }: ICheckColumnsCombinations) => {
+  let isMismatch = false;
```
matedo1, line 167: `isMismatch` is doing actually nothing and can be removed

src/utils/utils.ts
```diff
@@ -148,3 +150,27 @@ export const calculateFileSize = (sizeInBytes: number): string => {
...
+  combinations.every((combination) => {
```
matedo1, line 168: if `true` and `false` are returned, then I would suggest following version should be used:

```tsx
return combinations.every((combination) => {
```

Then also `boolean` return type for the method needs to be added.

src/utils/utils.ts
```diff
@@ -148,3 +150,27 @@ export const calculateFileSize = (sizeInBytes: number): string => {
...
+    if (isMismatch || (combination.length >= 2 && combination.some((element) => !columns.includes(element)))) {
+      uiLogger.error(`Required combination unsatisfied: ${combinations} should be used as a combination.`);
```
matedo1, line 170: Use just singular `${combination}` to highlight which combination is failing even if it it's pointing only to first failing combination.

src/components/Username/Username.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { TooltipWrapper } from 'components/TooltipWrapper/TooltipWrapper';
+
+interface IUsernameProps {
+  text: string;
+  length?: number;
+}
+
+/**
+ * The username component that will display a truncated username.
```
matedo1, line 11:
```
The username component that will display a adjusted username. 

This component needs to be used always when username is displayed to keep consistency on all places.
```

matedo1 (changes requested):

src/utils/utils.ts
DnsZhou: The reason for using `every` instead of `forEach` is that `every` will terminate the iteration when first `false` is returned, and either `true` or `false` is required to be returned in each iteration. See doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

src/components/TruncatedText/TruncatedText.tsx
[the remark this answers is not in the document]
matedo1: done

src/utils/utils.ts
matedo1: Using `every` is fine, also `true` and `false`, but I was meaning also returning it would be more usual. If you prefer your void based version, let's keep it, we can change it later when required.

src/components/TruncatedUsername/TruncatedUsername.tsx
```diff
@@ -0,0 +1,28 @@
...
+  length?: number;
+}
+
+/**
+ * The username component that will display a truncated username.
+ *
+ * @param text - The raw text of the username to be truncated
+ * @param length - The target length for usernames that are too long, by default 20
+ */
+export const TruncatedUsername = ({ text, length = 20 }: ITruncatedTextProps) => {
```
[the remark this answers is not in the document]
matedo1: done

src/utils/utils.ts
matedo1: done

src/components/TruncatedUsername/TruncatedUsername.tsx
matedo1: done

src/utils/utils.ts
matedo1: done

src/utils/utils.ts
matedo1: done

src/components/Username/Username.tsx
matedo1: done

matedo1 (approved):
There is last one optional proposal pending, but it's approved.

Entity Attributes were already integrated into Sorting, rebase your PR on top of it and apply required changes, see for example:

- https://github.com/project-ncl/pnc-web-ui-react/pull/273/files#diff-9be327f02fe34e9110d5a73d2f501fd115e816052eb6461d21d8e36160a7abff
- https://github.com/project-ncl/pnc-web-ui-react/pull/273/files#diff-bbedf30c67e0ca2f1d9846bf98525f1294244dd631536053e131d344787c0dc3

Thanks!

src/utils/utils.ts
DnsZhou: Updated accordingly

DnsZhou:
Hi @matedo1 , I updated the code to align the latest upstream, see the latest commit and feel free to approve it if it makes sense. Thanks!

matedo1 (approved):

PR #262, opened 2023-07-26
[NCL-7223] Integrate Group Configs list on Version pages

patrikk0123:


DnsZhou (approved):
Verified from my local env without console error, functionally working good, Thanks.

PR #275, opened 2023-08-09
NCL-8122 Refactor getFilteringAttributes and apply useMemo

matedo1:


DnsZhou:
When introducing/removing `useMemo` hooks to the component, **sometimes** there are some impacts to the rendering process thus causing the Unit Test snapshots to be changed(some attributes like `aria-labelledby` of the button component would be changed slightly).

Thus we need to pay more attention to the snapshot changes after introducing `useMemo` hooks into a component, snapshot tests from any pages that use this component may need to be updated.

* Be aware that this snapshot change may not happen every time, but do check unit-test results when making any changes over `useMemo`.

See [example snapshot changes](https://github.com/project-ncl/pnc-web-ui-react/pull/275/commits/66c09fbb0495fffcb4a9b22825db6aaff8d69f70#diff-7474bb6634e064d055791ec8875f2b7be96129eb66fa6773f69140e3ed320bba) that caused by introducing `useMemo` hooks.

patrikk0123 (commented):
I approve the PR

matedo1:
Great, thanks

PR #277, opened 2023-08-10
NCL-7212 Implement Group Builds page

matedo1:


src/components/GroupBuildsList/GroupBuildsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { Toolbar } from 'components/Toolbar/Toolbar';
+import { Username } from 'components/Username/Username';
+
+import { areDatesEqual, calculateDuration, createDateTime } from 'utils/utils';
+
+interface IGroupBuildsListProps {
+  serviceContainerGroupBuilds: IServiceContainer;
+  componentId: string;
+}
+
```
DnsZhou, line 36: I would suggest also implementing `defaultColumns` feature since columns would be different on other pages(like the homepage dashboards that have no users column)

src/components/GroupBuildsPage/GroupBuildsPage.tsx
```diff
@@ -1,23 +1,32 @@
...
+import { useServiceContainer } from 'hooks/useServiceContainer';
 import { useTitle } from 'hooks/useTitle';
 
-export const GroupBuildsPage = () => {
+import { GroupBuildsList } from 'components/GroupBuildsList/GroupBuildsList';
+import { PageLayout } from 'components/PageLayout/PageLayout';
+
+import * as groupBuildApi from 'services/groupBuildApi';
+
+export const GroupBuildsPage = ({ componentId = 'g1' }) => {
```
DnsZhou, line 12: I am also using `g1` for group config pages, but maybe I need to change to `gc1`, shall we have some naming rules for componentIds?

DnsZhou (changes requested):
Hi @matedo1 , it looks all good and verified on my local. From the code side, maybe we also need defaultColumns feature for this one since the groupBuildList would be used on the home page dashboard as well.

src/components/GroupBuildsList/GroupBuildsList.tsx
matedo1: done

src/components/GroupBuildsPage/GroupBuildsPage.tsx
matedo1: I changed Group Builds page id to `gb1`, please change also Group Configs page id to `gc1` to avoid some initial conflicts.

matedo1:
Hi @DnsZhou , changes done, could you do a code-review for me?

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -27,13 +27,15 @@ import { Username } from 'components/Username/Username';
 
 import { areDatesEqual, checkColumnsCombinations, createDateTime } from 'utils/utils';
 
+type TColumns = Array<keyof typeof buildConfigEntityAttributes>;
+
 interface IBuildConfigsListProps {
   serviceContainerBuildConfigs: IServiceContainer;
-  columns?: Array<keyof typeof buildConfigEntityAttributes>;
+  columns?: TColumns;
```
DnsZhou, line 34: Shall we also change it for the artifacts list to make it universal? https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/ArtifactsList/ArtifactsList.tsx#L44

DnsZhou (changes requested):
Looks good and verified locally, just one thing about the columns. Thanks!

src/components/BuildConfigsList/BuildConfigsList.tsx
matedo1: Yes, I know about this one, can you create separated jira for that?

PR #280, opened 2023-08-16
[NCL-7227] Implement Product Milestone Close action

patrikk0123:


DnsZhou (approved):
All features are verified from my local, and the code looks fine to me. Consider putting Jira number(or create a new Jira if not yet created) in todos to help track the status of tasks to be done in the future.

PR #287, opened 2023-08-22
[NCL-8130] Remove scrollbar when topbar is displayed

patrikk0123:


DnsZhou (approved):
The code looks good to me and was verified from my local, just add TSDocs for the new hook, Thanks.

PR #288, opened 2023-08-22
[NCL-8111] Make uiLogger not send localhost logs

patrikk0123:


src/services/uiLogger.ts
```diff
@@ -71,6 +71,8 @@ const createData = (message?: string, error?: Error, additionalData?: Object): I
 };
 
 const log = (message: string, error?: Error, additionalData?: Object) => {
+  if (window.location.hostname === 'localhost') return;
```
matedo1, line 74: For local development we still need some logs to be visible, so that we are aware there is some issue

matedo1 (changes requested):

matedo1 (approved):

PR #289, opened 2023-08-23
[NCL-8135] Create extendRequestConfig util function

patrikk0123:


src/utils/utils.ts
```diff
@@ -181,3 +186,29 @@ export const checkColumnsCombinations = ({ columns, combinations }: ICheckColumn
     }
   });
 };
+
```
matedo1, line 189: I would create `utils/requestConfigHelper.ts` and put all related code to this to make it clear what code parts are meant to be together.

src/utils/utils.ts
```diff
@@ -181,3 +186,29 @@ export const checkColumnsCombinations = ({ columns, combinations }: ICheckColumn
...
+type TExtendString = string & { _excludeType: 'q' };
+
+interface IExtendQParam {
```
matedo1, line 192: This should be probably just `IQParam`

src/utils/utils.ts
```diff
@@ -181,3 +186,29 @@ export const checkColumnsCombinations = ({ columns, combinations }: ICheckColumn
...
+  id: string;
+  value: string;
+  operator: IQParamOperators;
+}
+
+interface IExtendParams {
+  qParams?: IExtendQParam[];
+  [key: TExtendString]: string;
+}
+
```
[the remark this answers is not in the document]
matedo1: I would expect that I should be able to use for example:
```tsx
const newReqestConfig = extendRequestConfig(requestConfig, { timeout: 123 });
```

but it will result in:
```
Argument of type '{ timeout: number; }' is not assignable to parameter of type 'IExtendParams'.
  Object literal may only specify known properties, and 'timeout' does not exist in type 'IExtendParams'.
```

matedo1 (changes requested):
I would propose to start with demo meeting before any changes are done. I have several questions to this. Thanks for investigating this!

PR #290, opened 2023-08-23
[NCL-7775] Rework calculateDuration

patrikk0123:


DnsZhou (approved):
Verified from my local, it looks better now. Thanks @patrikk0123

PR #296, opened 2023-09-04
[NCL-7222] Implement Product create/edit page

patrikk0123:


src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const formComponent = (
+    <ContentBox padding isResponsive>
+      <Form
+        onSubmit={(e) => {
+          e.preventDefault();
+        }}
+      >
+        <FormGroup
+          isRequired
```
matedo1, line 103: [optional] required information needs to be passed on 3 different places:
1) formConfig
2) FormGroup
3) Input

Is it possible to propagate it from formConfig to other places automatically?

src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+            </FormHelperText>
+          }
+        >
+          <TextInput
+            isRequired
+            type="text"
+            id={productEntityAttributes.name.id}
+            name={productEntityAttributes.name.id}
+            autoComplete="off"
+            {...register<string>(productEntityAttributes.name.id, formConfig.name)}
```
matedo1, line 118: [optional]
```tsx
...register<string>(productEntityAttributes.name.id, formConfig.name)
```
could be changed to:
```tsx
... = useForm(formConfig);
// ...
...register<string>(productEntityAttributes.name.id)
```

It means `formConfig.name` would be automatically derived from `productEntityAttributes.name.id`, but then `formConfig` 1-level attribute keys would have to match entity attribute, TypeScript checker could be one of the possible solutions.

src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+      })
+      .then((response: any) => {
+        const newProductId = response?.data?.id;
+        if (!newProductId) {
+          throw new Error(`Invalid productId coming from Orch POST response: ${newProductId}`);
+        }
+        navigate(`/products/${newProductId}`);
+      })
+      .catch((e: any) => {
+        throw new Error('Failed to create Product.');
```
matedo1, line 70: Can you please make sure if errors like this are handled properly? Feel free to create new jira and do some additonal investigation. We need to be sure:
1) user knows what he is supposed to do when error happends
2) all errors which should be logged are logged

matedo1 (approved):
I put some not directly related questions / ideas, but this PR is approved, thanks.

src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
patrikk0123: Won't do, useForm hook will not return props specific to PatternFly components.

matedo1: I don't feel it PatternFly specific, you could say exactly the same about other properties that are returned at this moment. Even if we use own input component or other 3rd party component, some `isRequired` flag would be still needed to prevent duplications. 

`useForm` is managing validations -> `isRequired` is directly related to validations, so also from this point of view, it's part of its scope.

The thing is that at this moment it's easy to create some inconsistency, if any of the `isRequired` is missing by mistake or there can be some unwanted left overs.

matedo1: We agreed it won't be implemented now, this decision can be revalidated again later.

src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
matedo1: I created new jiras NCL-8169 and NCL-8170

src/components/ProductCreateEditPage/ProductCreateEditPage.tsx
matedo1: Hi @patrikk0123 , can you mention here what was the final decision?

PR #305, opened 2023-09-12
[NCL-8161] Add validation to input fields on Project pages

patrikk0123:


src/components/ProjectCreateEditPage/ProjectCreateEditPage.tsx
```diff
@@ -14,19 +14,29 @@ import { ServiceContainerCreatingUpdating } from 'components/ServiceContainers/S
 
 import * as projectApi from 'services/projectApi';
 
-import { validateUrl } from 'utils/formValidationHelpers';
+import { maxLength, validateUrl } from 'utils/formValidationHelpers';
 import { createSafePatch } from 'utils/patchHelper';
 import { generatePageTitle } from 'utils/titleHelper';
 
+const urlValidator = { validator: validateUrl, errorMessage: 'Invalid URL format.' };
+const maxLengthValidator = { validator: maxLength(255), errorMessage: 'Maximal input length is 255 characters.' };
```
matedo1, line 22: I guess maxLengthValidator , especially length 255 will be repeated over and over, so there are two improvements:
1) just move and export this variable, I would probably rename it to `maxLengthValidator255` (next there can be `maxLengthValidator30`)  so that also the other places will be able to use it = we will be sure all our error messages will be consistent 
2) move and export new function `maxLengthValidator(255)`, new preconfigured object with `255` would be returned

matedo1: I would say we could do also something similar for the `urlValidator` as it won't be so often, but still quite usual

matedo1 (changes requested):

matedo1 (approved):

PR #302, opened 2023-09-07
[NCL-7213] Group Build Detail Page

DnsZhou:

PR #321, opened 2023-10-11
[NCL-8173] Implement nested tabs

patrikk0123:


patrikk0123:
Will be applied on Group Build detail page once https://github.com/project-ncl/pnc-web-ui-react/pull/302 is merged.

src/components/NestedTabs/NestedTabs.tsx
```diff
@@ -0,0 +1,13 @@
+import { Tabs } from '@patternfly/react-core';
+import { PropsWithChildren } from 'react';
+
+interface INestedTabsProps {
+  activeTabKey: number;
+  onTabSelect: (tabKey: INestedTabsProps['activeTabKey']) => void;
+}
+
+export const NestedTabs = ({ activeTabKey, onTabSelect, children }: PropsWithChildren<INestedTabsProps>) => (
```
matedo1, line 9: Add also TSDoc. It has to include also its recommended usage - it can be used only in very specific cases (when switch-like functionality is sufficient) as we require all pages to be reachable via link, but in this case it's not fulfilled. For this purpose there is another jira, but we won't implement it now.

matedo1 (changes requested):

matedo1:
Hi @DnsZhou , can you please do a code-review for this one?

DnsZhou (approved):
So far it looks good to me, I think this should be merged before NCL-7213(https://github.com/project-ncl/pnc-web-ui-react/pull/302) so that I can implement this in 7213.

PR #325, opened 2023-10-17
[NCL-7812] Upgrade BrowserRouter to createBrowserRouter (v6.4)

jomrazek:


matedo1:
Hi @patrikk0123 , for now we agreed `createRoutesFromElements` can be used so that we can keep current declarative structure. In the future we can consider migrating it to object format. Or do you have any preferences?

src/AppRoutes.tsx
```diff
@@ -60,8 +61,8 @@ import { VariablesPage } from 'components/VariablesPage/VariablesPage';
 
 import { AUTH_ROLE } from 'services/keycloakService';
 
-export const AppRoutes = () => (
-  <Routes>
+export const AppRoutes = createRoutesFromElements(
```
patrikk0123, line 64: I would rather apply the function where the `createBrowserRouter` is called.
This way, AppRoutes remains a component.

patrikk0123:
> Hi @patrikk0123 , for now we agreed `createRoutesFromElements` can be used so that we can keep current declarative structure. In the future we can consider migrating it to object format. Or do you have any preferences?

Using `createRoutesFromElements` is the right choice. The declarative syntax is better and functionally, both ways seem equivalent.

src/AppRoutes.tsx
jomrazek: sure :+1:

patrikk0123 (approved):

PR #328, opened 2023-10-19
NCL-8135 Extend request config

matedo1:


matedo1:
Hi @jomrazek , when testing it, please make sure
 - `originalParams` are more complicated = use filtering using `q` param, enable `buildConfigName` filtering on Build page (see `BuildsList.tsx > defaultColumns`)
 - try to create more advanced `newParams` that will contain for example some 
    - protected param (like `pageSize`) 
    - conflict param (for example with `buildConfigName` I mentioned previously)
    - new param

jomrazek (approved):
Looks ok to me, also I tested it generally and for correct behavior in corner cases, works ok.

PR #329, opened 2023-10-23
[NCL-8129] [React] Fix icons size

jomrazek:


src/components/IconWrapper/IconWrapper.tsx
```diff
@@ -0,0 +1,16 @@
...
+ * Icon Wrapper aims to set minimal width to icons and stop them disapearing with resizing.
+ *
+ * @param children - icon to display
+ */
+export const IconWrapper = ({ children }: React.PropsWithChildren<IIconWrapper>) => {
+  return (
+    <span className={styles['icon-wrapper']}>
+      <>{children}</>
+    </span>
+  );
```
patrikk0123, line 15: - `return` statement is not needed (`() => <span>...</span>` syntax is more suitable here)
- Import `PropsWithChildren` directly
- Use `{children}`(without the `<></>`)

jomrazek: Thank you

patrikk0123 (approved):

PR #332, opened 2023-10-30
[NCL-8247] Refactor Dependency Tree component

patrikk0123:


patrikk0123 (commented):
`defaultExpanded` is not an official way to keep the nesting level and that causes problems.

For example, you cannot go from nesting level 2 to nesting level 1 (although the opposite works), because the TreeView component does not get properly re-rendered.

@DnsZhou in the old implementation, you made a workaround by the `refreshComponent` function. Another workaround is to set the all-expanded state to false temporarily. But these workarounds cause flickering.

DnsZhou (approved):
All feature have been verified working, and PF Tree workaround seems good within 20 levels of builds. for more than 20 builds it works exactly as what PF Tree component looks like, which is fine. Thanks for the fix!

PR #333, opened 2023-11-01
[NCL-8117] Implement Build Brew Push page

patrikk0123:


patrikk0123 (commented):
Live log was separated to a Jira NCL-8268, since the websockets still were not integrated.

DnsZhou (changes requested):
Hi @patrikk0123 , the code mostly looks good but I found that when I hit the `push to brew` button on the right top of the brew push page and proceed, the `Open brew push details` button does not work as expected. I have to click close and refresh button to see the result. is this an expected behaviour or we can improve it? ideally it should be refreshed automatically when brew push was proceeded.

matedo1:
Hi @DnsZhou we keep it open and display status (pending, success, error) there. Once finished, user can decide what action should be performed next (refresh, go to the detail + refresh, etc). Different modal windows can have different action, but we want to keep the behavior the same.

DnsZhou (approved):

PR #341, opened 2023-11-13
[NCL-7205] Build Config details page - main structure and details

DnsZhou:


patrikk0123:
Fix the errors pipeline is failing for first.

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+            <AttributesItem title={buildConfigEntityAttributes.modificationUser.title}>
+              {serviceContainerBuildConfig.data?.modificationUser?.username}
+            </AttributesItem>
+
+            <AttributesItem title={buildConfigEntityAttributes.description.title}>
+              {serviceContainerBuildConfig.data?.description}
+            </AttributesItem>
+
+            <AttributesItem title={buildConfigEntityAttributes.buildType.title}>
+              {serviceContainerBuildConfig.data?.buildType}
```
patrikk0123, line 112: `BuildConfigBuildTypeLabelMapper` exists

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+      firstLevelEntity: 'Build Config',
+      entityName: serviceContainerBuildConfig.data?.identifier,
+    })
+  );
+
+  return (
+    <Grid hasGutter>
+      <GridItem span={12}>
+        <ContentBox padding marginBottom isResponsive>
+          <Attributes>
```
patrikk0123, line 88: Empty lines between `AttributesItem`s are placed inconsistently.

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <CodeBlock>
+                <CodeBlockCode> {serviceContainerBuildConfig.data?.buildScript}</CodeBlockCode>
+              </CodeBlock>
+            </AttributesItem>
+            <AttributesItem title={buildConfigEntityAttributes.brewPullActive.title}>
+              {serviceContainerBuildConfig.data?.brewPullActive + ''}
+            </AttributesItem>
+            <AttributesItem title={buildConfigEntityAttributes.productVersion.title}>
+              {serviceContainerBuildConfig.data?.productVersion && (
+                <ProductVersionLink productVersion={serviceContainerBuildConfig.data?.productVersion} />
```
patrikk0123, line 136: `serviceContainerBuildConfig.data.productVersion`

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  key={index}
+                  title={parameterKey}
+                  tooltip={
+                    parameterKey === 'ALIGNMENT_PARAMETERS' ? (
+                      <>
+                        See{' '}
+                        <Link to="http://release-engineering.github.io/pom-manipulation-ext/#feature-guide">Feature Guide</Link>
+                      </>
+                    ) : (
+                      ''
```
patrikk0123, line 164: prefer `undefined`

src/components/Attributes/AttributesItem.tsx
```diff
@@ -9,7 +9,7 @@ import styles from './Attributes.module.css';
 
 interface IAttributesItemProps {
   title: React.ReactNode;
-  tooltip?: string;
+  tooltip?: string | ReactNode;
```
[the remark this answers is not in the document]
DnsZhou: I would like to keep it simple so will keep using ReactNode, thx.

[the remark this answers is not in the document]
DnsZhou: Also this is just keeping the same with what React Tooltip component do, which is accepting `ReactNode` type as the tooltip.

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -11,8 +18,31 @@ interface IProductVersionLinkProps {
...
+    if (productVersion && !productVersion.product) {
+      serviceContainerProductVersionRunner({ serviceData: { id: productVersion.id } });
+    }
+  }, [serviceContainerProductVersionRunner, productVersion]);
+
+  return (
+    <>
+      {productVersion && productVersion.product && (
+        <Link to={`/products/${productVersion.product?.id}/versions/${productVersion.id}`}>
+          {productVersion.product?.name} - {productVersion.version}
```
patrikk0123, line 35: `{productVersion.product.name}`

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -11,8 +18,31 @@ interface IProductVersionLinkProps {
...
+  useEffect(() => {
+    if (productVersion && !productVersion.product) {
+      serviceContainerProductVersionRunner({ serviceData: { id: productVersion.id } });
+    }
+  }, [serviceContainerProductVersionRunner, productVersion]);
+
+  return (
+    <>
+      {productVersion && productVersion.product && (
+        <Link to={`/products/${productVersion.product?.id}/versions/${productVersion.id}`}>
```
patrikk0123, line 34: `${productVersion.product.id}`

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -11,8 +18,31 @@ interface IProductVersionLinkProps {
...
  * @param productVersion - Product Version object
  */
-export const ProductVersionLink = ({ productVersion }: IProductVersionLinkProps) => (
-  <Link to={`/products/${productVersion.product?.id}/versions/${productVersion.id}`}>
-    {productVersion.product?.name} - {productVersion.version}
-  </Link>
-);
+export const ProductVersionLink = ({ productVersion }: IProductVersionLinkProps) => {
+  const serviceContainerProductVersion = useServiceContainer(productVersionApi.getProductVersion);
+  const serviceContainerProductVersionRunner = serviceContainerProductVersion.run;
```
patrikk0123, line 23: I understand that Build Configs provides only `ProductVersionRef` not containing the Product name and ID, but this component needs to remain dumb, therefore, the Version should be fetched one level up. The behavior is really weird (it accepts a Product Version and fetches the same Product Version) - even if one is a Ref and the other a normal Version, it is not right.

DnsZhou: If we want the product fetching process level-up to where we use this component, then the developer who uses the component will need to know whether the productVersion they passed to the component actually has a product object. This is not a good practice in terms of the encapsulation concept, so what I will do is to refactor the ProductVersionLink component to just accept the productVersionLink which will make it look better.

patrikk0123: That's not the best choice either, because the Version Link component should parse and create the link, not accept it (if that is what you meant).

A better solution is to ensure the component can accept only Version, where Product IS defined:
```tsx
interface ProductVersionWithProduct extends ProductVersion {
  product: Product;
}

interface IProductVersionLinkProps {
  productVersion: ProductVersionWithProduct;
}
```

or

```ts
interface IProductVersionLinkProps {
  productVersionId: string;
  productId: string;
}
```

patrikk0123: btw we right now use both approaches on our link components, some accept IDs and some entity objects, which is quite unfortunately inconsistent. @matedo1

src/common/buildConfigEntityAttributes.ts
```diff
@@ -66,4 +66,40 @@ export const buildConfigEntityAttributes = {
...
+    id: 'environment',
+    title: 'Environment',
+  },
+  scmRepository: {
+    id: 'scmRepository',
+    title: 'SCM Repository',
+  },
+  scmRevision: {
+    id: 'scmRevision',
+    title: 'Revision',
```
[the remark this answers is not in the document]
DnsZhou: Fixed

src/components/ProductVersionLink/ProductVersionLink.tsx
DnsZhou: Hi both,

Still, I am not very convinced to accept redundant data that actually can be enough with one id. Our GET productVersion endpoint actually returns a ProductVersion object that contains also the Product object in it, which means taking only productVersion ID is enough, if we are accepting both id then what's the point of that? 

Thanks.
cc @patrikk0123 @matedo1

DnsZhou: Also just a reminder that we are not only parsing the URL to the product version but also displaying the product name and version name here, thus only the product id and productVersion id are not enough, we do need at least these names.

I do agree that a dumb component can be better, especially when maybe we are going to display a lot of product version links in a list. But given the situation that our backend sometimes can provide not a full version object but instead a ref version of the object (which is the root cause of why we had this discussion: there is no product information in the `ProductVersion` object in the `BuildConfig` entity) (see the long-history discussion of NCL-8124), seems at this moment we have no choice but fire a request to get these information we need(productVersion name, product id, product name).

If we have to send a request to the backend, I would suggest using my original implementation, which is  still accepting the productVersion data, but inside the component, if we find there is no product entity in the entity we then fire a request to get the ProductVersion entity. The good reason for doing this is that 1). other developers do not need to worry about whether this is a regular version of ProductVersion entity or ProductVersionRef that has no product entity, and 2).especially in the future we wish to totally remove the `ref` version of entity, and by that time it will be easy to just change the productVersion component to a dumb one without changing anywhere else.

I will revert this component back to the version I mentioned above first, and we can discuss further over that.

patrikk0123: You are right that the component also displays the title, so instead of passing IDs of Product and Version, I would pass actual entities Product and Version to the component. If the link is used with the Version entity, the Product in it can be passed, if it's VersionRef, the Product needs to be fetched in the component rendering the link.

So of course, if these required data (Product) are not available, they need to be fetched, but this component has to remain dumb so the fetch will be done one level up.

Also, the component might defaultly get the Procuct from the passed Version, like this
```ts
const ProductVersionLink = ({
  productVersion,
  product = productVersion.product
}) => ...
```

src/components/GroupConfigDetailPage/GroupConfigDetailPage.tsx
```diff
@@ -113,7 +113,9 @@ export const GroupConfigDetailPage = ({
                 variant="inline"
                 title={groupConfigEntityAttributes.productVersion.title}
               >
-                <ProductVersionLink productVersion={serviceContainerProductVersion.data!} />
+                {serviceContainerProductVersion.data && (
+                  <ProductVersionLink productVersion={serviceContainerProductVersion.data} />
+                )}
```
patrikk0123, line 118:
```
<ProductVersionLink productVersion={serviceContainerProductVersion.data!} />
```
the data are children of `ServiceContainerLoading`

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const serviceContainerProductVersion = useServiceContainer(productVersionApi.getProductVersion);
+  const serviceContainerProductVersionRunner = serviceContainerProductVersion.run;
+
+  const tootipLinkStyle: CSSProperties = {
+    color: '#55AAFF',
+  };
+
+  useEffect(() => {
+    if (serviceContainerBuildConfig.data?.productVersion) {
+      serviceContainerProductVersionRunner({ serviceData: { id: serviceContainerBuildConfig.data!.productVersion!.id } });
```
patrikk0123, line 37: `serviceContainerBuildConfig.data.productVersion.id`

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+            </AttributesItem>
+            <AttributesItem title={buildConfigEntityAttributes.productVersion.title}>
+              <ServiceContainerLoading
+                {...serviceContainerProductVersion}
+                variant="inline"
+                title={buildConfigEntityAttributes.productVersion.title}
+              >
+                {serviceContainerProductVersion.data && (
+                  <ProductVersionLink productVersion={serviceContainerProductVersion.data} />
+                )}
```
patrikk0123, line 106:
```
<ProductVersionLink productVersion={serviceContainerProductVersion.data!} />
```

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -1,18 +1,24 @@
...
  */
-export const ProductVersionLink = ({ productVersion }: IProductVersionLinkProps) => (
-  <Link to={`/products/${productVersion.product?.id}/versions/${productVersion.id}`}>
-    {productVersion.product?.name} - {productVersion.version}
-  </Link>
-);
+export const ProductVersionLink = ({ productVersion, product = productVersion.product }: IProductVersionLinkProps) => {
+  !product && uiLogger.error(`product has to be defined in ProductVersionLink component`);
+  return (
+    <Link to={`/products/${product?.id}/versions/${productVersion?.id}`}>
```
patrikk0123, line 20: `productVersion.id`

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -1,18 +1,24 @@
...
  *
  * @param productVersion - Product Version object
  */
-export const ProductVersionLink = ({ productVersion }: IProductVersionLinkProps) => (
-  <Link to={`/products/${productVersion.product?.id}/versions/${productVersion.id}`}>
-    {productVersion.product?.name} - {productVersion.version}
-  </Link>
-);
+export const ProductVersionLink = ({ productVersion, product = productVersion.product }: IProductVersionLinkProps) => {
+  !product && uiLogger.error(`product has to be defined in ProductVersionLink component`);
```
patrikk0123, line 18: throw an error, I think there's no need to log it

patrikk0123 (commented):
Also merge the commits so there are three in total (build pages structure, build detail page and tooltip change).

DnsZhou:
Hi @patrikk0123 , feel free to check it, Thanks.

src/components/GroupConfigDetailPage/GroupConfigDetailPage.tsx
patrikk0123: what I meant is that `serviceContainerProductVersion.data` check is not needed

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -1,18 +1,25 @@
...
-    {productVersion.product?.name} - {productVersion.version}
-  </Link>
-);
+export const ProductVersionLink = ({ productVersion, product = productVersion.product }: IProductVersionLinkProps) => {
+  if (!product) {
+    throw new Error('product has to be defined in ProductVersionLink component');
+  }
+
+  return (
+    <Link to={`/products/${product?.id}/versions/${productVersion.id}`}>
```
patrikk0123, line 21: `product.id`

src/components/ProductVersionLink/ProductVersionLink.tsx
```diff
@@ -1,18 +1,25 @@
...
+      {product?.name} - {productVersion.version}
```
patrikk0123, line 22: `product.name`

src/components/BuildConfigDetailPage/BuildConfigDetailPage.tsx
patrikk0123: what I meant is that serviceContainerProductVersion.data check is not needed

patrikk0123 (approved):

PR #346, opened 2023-11-27
[NCL-7207] Implement Build Config create/edit page - general attributes section

patrikk0123:


src/services/environmentApi.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { pncClient } from './pncClient';
+
+export interface IEnvironmentApiData {
+  id: string;
+}
+
+/**
+ * Gets all Environments.
+ *
```
matedo1, line 15: Missing TSDoc for `hidden` and `deprecated` parameters and their possible values:
 - `true` - description here
 - `false` - ...
 - `undefined` - ...

But I would consider more self-descriptive values (deprecated context):
 - `WITHOUT_DEPRECATED` (default) - ...
 - `ONLY_DEPRECATED` - ...
 - `ALL` - ...

or

 - `NOT_DEPRECATED_ONLY` (default) - ...
 - `DEPRECATED_ONLY` - ...
 - `ALL` - ...

or

 - `NO` (default) - ...
 - `YES_ONLY` - ...
 - `YES_ALSO` - ...

src/components/BuildConfigCreateEditPage/BuildConfigCreateEditPage.tsx
```diff
@@ -0,0 +1,345 @@
...
+                    setSelectedEnvironment(environment);
+                  }}
+                  onClear={() => {
+                    onChange('');
+                    setSelectedEnvironment(undefined);
+                  }}
+                  fetchCallback={fetchEnvironments}
+                  titleAttribute="description"
+                  placeholderText="Select Environment"
+                  getHighlightText={(environment: Environment) =>
```
matedo1, line 207: I think `getHighlightText` is too specific and there is high chance that there will be other requirements soon resulting in some other attributes which will be hard to maintain, to handle them more easily I would use general text modification instead, so that everyone can decide whether to use:
 - default text
 - custom text (if defined here)

Different use-cases may force you to use different "highlight" techniques, orders, etc. Using `customText` would be universal.

Example

```tsx
// ...
customText={(environment: Environemnt) => {
  if (environemt.deprecated) { ... }
  return <>`generating custom content here will cover also other possible use cases more easily`</>
}}
// ...
```

matedo1 (changes requested):

PR #360, opened 2024-01-09
[NCL-8342] Update GH actions/checkout to v4

patrikk0123:


matedo1 (approved):
`actions/setup-node@v3` will be updated in separated PR (not created yet) when also Node + NPM is updated

PR #362, opened 2024-01-11
[NCL-7217] Fix empty-version error for groupConfigCreateEditPage

DnsZhou:


src/components/GroupConfigCreateEditPage/GroupConfigCreateEditPage.tsx
```diff
@@ -125,6 +129,7 @@ export const GroupConfigCreateEditPage = ({ isEditPage = false }: IGroupConfigCr
         const groupConfig: GroupConfiguration = response.data;
 
         if (groupConfig.productVersion) {
+          setFieldValues({ name: groupConfig.name });
```
patrikk0123, line 132: also please remove setting of `name` in the `setFieldValues` call below since it is set here now

src/components/GroupConfigCreateEditPage/GroupConfigCreateEditPage.tsx
```diff
@@ -97,7 +98,10 @@ export const GroupConfigCreateEditPage = ({ isEditPage = false }: IGroupConfigCr
   };
 
   const submitEdit = (data: IFieldValues) => {
-    const newData = { name: data.name, productVersion: { id: selectedProductVersion!.id } };
+    const newData = {
+      name: data.name,
+      productVersion: selectedProductVersion ? ({ id: selectedProductVersion?.id } as ProductVersion) : null,
```
patrikk0123, line 103: `selectedProductVersion ? ({ id: selectedProductVersion.id }...`

patrikk0123 (commented):
Also please rename the commit to more appropriate name

src/components/GroupConfigCreateEditPage/GroupConfigCreateEditPage.tsx
DnsZhou: image
unfortunately the `setFieldValues` function, as how it was named, will override any existing value thus causing a result like this. So we have to keep it. This also troubles me a lot because sometimes I do want to change specific fields, are we planning to have `setFieldValue` function for useForm hook in the future?

patrikk0123: Okay, I did not realize this.
`setFieldValue` might be implemented.

DnsZhou:
Hi @patrikk0123 , could you help review this pr? Thanks

patrikk0123 (approved):

PR #372, opened 2024-01-29
NCL-7459 Implement WebSocket client

matedo1:


src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+import * as pncWebSocketService from 'services/pncWebSocketService';
+
+/**
+ * HOOK CHECKERS
+ */
+
+/**
+ * Additional filtering parameters
+ */
+interface IOnParameters {
```
patrikk0123, line 12: `IOnParameters` sound general but these parameters seem to be related to Build process.

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * Check whether Build status change WebSocket event was sent.
+ *
+ * @param wsData - WebSocket data
+ * @param parameters - See {@link IOnParameters}
+ * @returns true when Build status changed, otherwise false
+ */
+export const onBuildStatusChanged = (wsData: any, { buildConfigId, userId }: IOnParameters = {}) => {
```
patrikk0123, line 24: `on*` naming convention is used for event handlers (accepting and executing callback on event). This function just returns a boolean (also, the return type is missing). So maybe callback should be passed?

patrikk0123, line 24: `wsData` is any

patrikk0123, line 24: extract `on*` functions to one separate file

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+  let result = wsData.job === 'BUILD' && wsData.notificationType === 'BUILD_STATUS_CHANGED';
+  if (result && buildConfigId) {
+    result = result && wsData.build.buildConfigRevision.id === buildConfigId;
+  }
+  if (result && userId) {
+    result = result && wsData.build.user.id === userId;
+  }
+  return result;
```
patrikk0123, line 32: `let` can be avoided. Like this, for example.

```ts
const result = wsData.job === 'BUILD' && wsData.notificationType === 'BUILD_STATUS_CHANGED';
return result &&
            (!buildConfigId || wsData.build.buildConfigRevision.id === buildConfigId) &&
            (!userId || wsData.build.user.id === userId);
```

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * HOOK
+ */
+
+interface IUsePncWebSocketEffectOptions {
+  // condition based listening, when true, events will be ignored (for example when Build is finished)
+  preventListening?: boolean;
+
+  // debug identification string
+  debug?: string;
```
patrikk0123, line 54: Remove this eventually.

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * WebSocket message listener hook.
+ *
+ * @param callback - Callback method, WebSocket parsed data attribute will be passed in
+ * @param options - See {@link IUsePncWebSocketEffectOptions}
+ */
+export const usePncWebSocketEffect = (
+  callback: Function,
```
patrikk0123, line 64: change it to more specific type

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+export const onBuildStarted = (wsData: any, { buildConfigId, userId }: IOnParameters = {}) =>
+  onBuildStatusChanged(wsData, { buildConfigId, userId }) && wsData.oldStatus === 'NEW';
+
+/**
+ * HOOK
+ */
+
+interface IUsePncWebSocketEffectOptions {
+  // condition based listening, when true, events will be ignored (for example when Build is finished)
+  preventListening?: boolean;
```
patrikk0123, line 51: This may be a misleading description when compared to implementation. The implementation stops callback from being registered. But when the condition changes, it will remain registered.

src/hooks/usePncWebSocket.ts
```diff
@@ -0,0 +1,81 @@
...
+        callback(wsData);
+      };
+
+      const removeMessageListener = pncWebSocketService.addMessageListener(onMessage, { debug });
+
+      return () => {
+        removeMessageListener();
+      };
+    }
+  }, [callback, preventListening, debug]);
```
patrikk0123, line 80: there might be a problem with the callback being registered on each re-render if it is not memoized

src/services/pncWebSocketService.ts
```diff
@@ -0,0 +1,54 @@
+import { pncWebSocketClient } from 'services/pncWebSocketClient';
+import { uiLogger } from 'services/uiLogger';
+
+type TWebSocketListener = (messageEvent: MessageEvent) => void;
+const messageListeners: Array<TWebSocketListener> = [];
+
+const LISTENERS_WARNING_COUNT = 20;
```
patrikk0123, line 7: maybe a little bit arbitrary number

src/services/pncWebSocketService.ts
```diff
@@ -0,0 +1,54 @@
+import { pncWebSocketClient } from 'services/pncWebSocketClient';
+import { uiLogger } from 'services/uiLogger';
+
+type TWebSocketListener = (messageEvent: MessageEvent) => void;
+const messageListeners: Array<TWebSocketListener> = [];
```
patrikk0123, line 5: Global state pollution. This is poorly extendable.
Generic websocket class having this state (and also listener add/remove methods) would be preferable.
Then, `pncWebSocketClient` would be an instance of that class.

src/services/pncWebSocketService.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const LISTENERS_WARNING_COUNT = 20;
+
+pncWebSocketClient.addEventListener('message', (message) => {
+  messageListeners.forEach((listener) => {
+    listener(message);
+  });
+});
```
patrikk0123, line 13: Direct dependency on one client. See another comment about global state pollution.

src/services/pncWebSocketService.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Add PNC WebSocket message listener.
+ *
+ * @param onMessageListener - Callback function processing WebSocket message
+ * @param debug - Debug identification string
+ * @returns removeMessageListener (unsubscribe) method
+ */
+export const addMessageListener = (onMessageListener: TWebSocketListener, { debug = '' }: IAddMessageListener = {}) => {
+  messageListeners.push(onMessageListener);
+  const random = Math.random(); // debug purposes
```
patrikk0123, line 28: remove this and other debugging lines

src/hooks/usePncWebSocket.ts
matedo1: `IOnBuildParameters`

src/hooks/usePncWebSocket.ts
matedo1: `has..`

src/hooks/usePncWebSocket.ts
matedo1: I will rename file to `usePncWebSocketEffect.ts`

src/hooks/usePncWebSocket.ts
matedo1: More advanced types are not available yet, but we will automatize them in the future.

src/hooks/usePncWebSocket.ts
matedo1: I created NCL-8377

src/hooks/usePncWebSocket.ts
matedo1: When the condition changes, `return` will be executed

src/hooks/usePncWebSocket.ts
matedo1: It has to be memoized.

src/services/pncWebSocketService.ts
matedo1: We can update it accordingly later once we have better data available.

src/hooks/usePncWebSocket.ts
matedo1: I renamed it.

patrikk0123 (approved):

PR #388, opened 2024-02-13
[NCL-8428] Integrate WebSockets on Close Result detail page

patrikk0123:


src/hooks/usePncWebSocketEffect.ts
```diff
@@ -190,3 +190,44 @@ export const hasBrewPushFinished = (wsData: any, { buildId }: IBrewPushParameter
 
   return !buildId || buildId === wsData.buildPushResult?.buildId;
 };
+
+/**
+ * Additional filtering parameters.
+ */
+interface IMilestonePushParameters {
```
matedo1, line 197: `IMilestoneCloseParameters`

matedo1 (changes requested):
Tab number update seems to be also missing. Thanks

matedo1 (approved):

PR #389, opened 2024-02-13
[NCL-8427] Integrate WebSockets on Deliverables Analysis detail page

patrikk0123:


matedo1 (approved):
NCL-8434 will provide additional possibilities to implement more advanced auto-refresh feature, but for now this the only way. Approved.

PR #390, opened 2024-02-14
[NCL-7205] Build Config revisions list and detail page

patrikk0123:


src/components/BuildConfigDetailPage/BuilConfigRevisionPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { BuildConfigDetail } from 'components/BuildConfigDetailPage/BuildConfigDetail';
+
+import * as buildConfigApi from 'services/buildConfigApi';
+
+export const BuildConfigRevisionPage = () => {
+  const { buildConfigId, revisionId } = useParamsRequired();
+
+  const serviceContainerBuildConfigRevision = useServiceContainer(buildConfigApi.getRevision);
+  const serviceContainerRevisionRunner = serviceContainerBuildConfigRevision.run;
```
matedo1, line 14: [optional] I would rather use full version `serviceContainerBuildConfigRevisionRunner`, it's useful when we are doing searching and expecting `serviceContainerBuildConfigRevision` part is included, you are using the full version also for example here: BuildConfigRevisionPages.tsx

matedo1 (approved):
Approved, just one minor thing, thanks

src/components/BuildConfigDetailPage/BuilConfigRevisionPage.tsx
patrikk0123: true, that was intended by me

PR #394, opened 2024-02-19
[NCL-8366] Integrate missing service URLs

patrikk0123:


matedo1 (approved):

matedo1:
We agreed to wait for backend fix first so that it can be fully tested before it's merged.

PR #396, opened 2024-02-19
[NCL-8357] Hide error on Action Modal when its data were changed

patrikk0123:


src/components/ActionModal/ActionModal.tsx
```diff
@@ -115,7 +115,10 @@ export const ActionModal = ({
     >
       {serviceContainer ? (
         <div className="m-r-0">
-          <ServiceContainerCreatingUpdating {...serviceContainer} title={actionTitle}>
+          <ServiceContainerCreatingUpdating
+            {...{ ...serviceContainer, error: wereSubmitDataChanged ? '' : serviceContainer.error }}
```
matedo1, line 119: I would prefer more flat and easy read structure like `<ServiceContainerCreatingUpdating {...serviceContainer} error={...} title={actionTitle}>`, the same structure is already used also on other similar places, but this one is also possible.

matedo1 (approved):

src/components/ActionModal/ActionModal.tsx
patrikk0123: That's right, I just did not realize it.

PR #404, opened 2024-02-29
[NCL-8456] Fix page titles

patrikk0123:


src/utils/titleHelper.ts
```diff
@@ -1,3 +1,5 @@
+import { isArray } from 'lodash';
```
matedo1, line 1: Once it's rebased, our function should be used.

matedo1 (approved):

PR #424, opened 2024-03-12
[NCL-8499] Fix Username component truncating

patrikk0123:


src/components/Username/Username.tsx
```diff
@@ -15,16 +17,27 @@ interface IUsernameProps {
  * @param length - The target length for usernames that are too long, by default 20
  */
 export const Username = ({ text, length = 20 }: IUsernameProps) => {
-  const removeKeyword = 'service-account-';
-  const resultText = text.replace(removeKeyword, '').substring(0, length);
+  const resultText = truncateUsername(text, length);
 
-  return text !== resultText ? (
+  return resultText.length !== text.length ? (
```
matedo1, line 22: This will display `..` automatically for all usernames containing `service-account-`, for example `service-account-a` will be displayed as `(SA) a..`

src/components/Username/Username.tsx
```diff
@@ -15,16 +17,27 @@ interface IUsernameProps {
...
     </TooltipWrapper>
   ) : (
     <>{text}</>
   );
 };
+
+const truncateUsername = (username: string, length: number): string => {
+  const truncatedUsername = username.replace(serviceAccountKeyword, '').substring(0, length);
+
+  // prettify the result
```
matedo1, line 37: [optional] `return truncatedUsername.replace(/-$/, '');` would be enough, or even more generalized regex `/-+$/`, so finally:

```ts
const truncateUsername = (username: string, length: number): string => username.replace(serviceAccountKeyword, '').substring(0, length).replace(/-+$/, '');
```

src/components/Username/Username.tsx
```diff
@@ -15,16 +17,27 @@ interface IUsernameProps {
...
-  const resultText = text.replace(removeKeyword, '').substring(0, length);
+  const resultText = truncateUsername(text, length);
 
-  return text !== resultText ? (
+  return resultText.length !== text.length ? (
     <TooltipWrapper tooltip={text}>
       <span>
-        {text.startsWith(removeKeyword) && <Label isCompact>SA</Label>} {resultText}
+        {text.startsWith(serviceAccountKeyword) && <Label isCompact>SA</Label>} {resultText}
+        <>..</>
```
matedo1, line 26: [optional] `<>` is not needed

src/components/Username/Username.tsx
```diff
@@ -2,6 +2,8 @@ import { Label } from '@patternfly/react-core';
 
 import { TooltipWrapper } from 'components/TooltipWrapper/TooltipWrapper';
 
+const serviceAccountKeyword = 'service-account-';
```
matedo1, line 5: [optional] If it's constant, then uppercase should be considered.

matedo1 (changes requested):

src/components/Username/Username.tsx
patrikk0123: No longer relevant (code changed).

PR #429, opened 2024-03-19
NCL-8507 Change Artifact Build layout

matedo1:


src/index.css
```diff
@@ -131,6 +131,10 @@ code {
   padding-right: 10px !important;
 }
 
+.p-r-20 {
+  padding-right: 20px !important;
+}
+
```
patrikk0123, line 137: `margin-right`, also don't put arbitrary gaps in index.css, instead, create `module.css` file and put it there.

matedo1: both options are possible, for simple cases when module.css is not needed generics seems to be reasonable too

PR #435, opened 2024-03-25
NCL-8387 Fix Product Version empty state on Build Config detail page

patrikk0123:


jomrazek (approved):
I don't know much about service container and where 0 argument should be used but other than that looks good

patrikk0123:
> I don't know much about service container and where 0 argument should be used but other than that looks good

It's there so the spinner is displayed immediately.

PR #450, opened 2024-04-03
[NCL-8591] Don't check for user role if authenticated

jomrazek:


thescouser89 (approved):
LGTM and nice catch!

Also, return those children!
```
return <>{children}</>;
```

matedo1 (approved):

PR #456, opened 2024-04-09
[NCL-8636] Use TextArea instead of CodeEditor on BC create/edit page

patrikk0123:


src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -8,6 +8,8 @@ interface ICodeEditorProps extends TextAreaProps {
   value: string;
 }
 
+// TODO: fix firefox/safari bug (NCL-8634)
```
matedo1, line 11: Please add also comment mentioning it's a dead code for now that it's not used at all

matedo1 (approved):

PR #465, opened 2024-04-12
NCL-8644 Add User support link

matedo1:


src/AppLayout.tsx
```diff
@@ -93,12 +101,29 @@ export const AppLayout = () => {
...
+      headerQuestionDropdownItems.push(
+        <DropdownItem key="users-guide" href={pncUserGuideUrl} target="_blank" rel="noopener noreferrer">
+          User's guide <ExternalLinkAltIcon />
+        </DropdownItem>
+      );
+    }
+
+    if (pncUserSupportUrl) {
+      headerQuestionDropdownItems.push(
+        <DropdownItem
```
matedo1, line 116: Here is the final button pointing to the User's support

cc: @pkocandr

jomrazek (approved):
Works nice, I have a feeling nobody will ever find it, but it is nice addition.

matedo1:
> Works nice, I have a feeling nobody will ever find it, but it is nice addition.

Other company projects use the same approach too, so our users might already be familiar with it. However, I am considering some other updates to make it more obvious. Thanks!

PR #470, opened 2024-04-18
[NCL-8657] Simplify switch labels

patrikk0123:


matedo1 (approved):
Thanks for looking into other similar cases

PR #481, opened 2024-04-22
[NCL-8671] Return missing Tooltip to Group Build status icon

patrikk0123:


matedo1 (approved):
Approved, 

anyway I've noticed we have already some similar component with the same name already: https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/components/BuildLogLink/BuildLogLink.tsx , we could try to merge them.

PR #482, opened 2024-04-22
[NCL-8418] Add Push-to-brew action to Group Build detail page

patrikk0123:


patrikk0123:
When Brew-push is accepted, no feedback is provided to the user regarding its success - unlike Build, the Group Build page has no Brew-push page, also, the latest Brew-push column in the Builds list is missing. But the user can see Brew-push result of individual Builds in the Group Build.

PR #486, opened 2024-04-24
[NCL-8675] Add latest Brew Push column to Group Build Builds list

patrikk0123:


patrikk0123:
Problems:
- There is no WS socket for Group Build push as a whole. Therefore, for each individual Build push WS message, the list is prefetched.
- Build push WS message contains no `groupBuildId` field

matedo1 (changes requested):

PR #487, opened 2024-04-24
Provide option for seconday url

jomrazek:


matedo1 (approved):
Approved, please update also Jenkins environments (devel, stage, prod) accordingly

PR #488, opened 2024-04-24
[NCL-8655] Refactoring SCM Url parse helper

DnsZhou:


DnsZhou:
Hi @matedo1 , I have changed to use env vars and all other parts remain the same. Please add GERRIT_URL_BASE and GERRIT_URL_STAGE_BASE to the env so that they can be recognized more accurately; otherwise, it will only use "code" as the keyword. anyway, this is just a temporary workaround and can be removed when we fully migrate to GitLab. Feel free to merge the code if you approve this change so that you can do the release as you mentioned.

DnsZhou:
I have also changed the test case and snapshots because the current test case matches the prefix, which has different behaviour between environments with env var and without env var(like github workflow env).

pkocandr:
IMO you don't  need 2 env variables for Gerrit. Based on the environment you are in you configure the GERRIT_URL_BASE value to the specific hostname, either prod or stage Gerrit.

DnsZhou:
> IMO you don't need 2 env variables for Gerrit. Based on the environment you are in you configure the GERRIT_URL_BASE value to the specific hostname, either prod or stage Gerrit.

That make sense, I missed this part. will change it. Thx! @pkocandr

patrikk0123 (approved):

PR #501, opened 2024-05-07
[NCL-8655] Fix port number issue for SCP URIs

DnsZhou:


src/utils/urlParseHelper.ts
```diff
@@ -25,9 +35,9 @@ export interface IParsedUrl {
 }
 
 /**
- * Parses SCM Repository URI or SCP to webview link of the SCM Repository.
+ * Parses SCM Repository URI or SCP-like URI to webview link of the SCM Repository.
  *
- * For SCP, the parser will ignore any port number and regard the text between ':' and '/' as
+ * For SCP-like URI, the parser will ignore any port number and regard the text between ':' and '/' as
```
patrikk0123, line 40: SCP-like URIs do not and cannot contain any port. It is normal URLS that can contain them.

patrikk0123 (approved):

src/utils/urlParseHelper.ts
patrikk0123: You probably meant:
```
In SCP-like URIs, the text after ':' is never regarded as the port but as part of the path...
```

DnsZhou: updated

PR #510, opened 2024-05-16
[NCL-8207] Fetch BC list on Project detail page parallelly  to  Project…

patrikk0123:
… details


matedo1 (approved):
In the future, we can probably make some improvements to the box size to avoid content jumping.

PR #513, opened 2024-05-20
[NCL-8543] Send only necessary payload on BC create/edit page

patrikk0123:


DnsZhou (approved):
lgtm, no issue with local test.

PR #515, opened 2024-05-21
[NCL-8703] Pre-build sync validation on the SCM edit page

patrikk0123:


DnsZhou (approved):
functionally verified from my local, and the new validator message & popup information looks good. code looks clear as well.

PR #516, opened 2024-05-22
[NCL-8678] Input Fields level Validaton

DnsZhou:


patrikk0123 (approved):
Code looks good, although I do not know whether the max lengths are accurate for all inputs.

PR #524, opened 2024-05-28
[NCL-8714] Add deprecated Environment filter to Build Config list component

patrikk0123:


DnsZhou (changes requested):
As the environment.deprecated should either be true or false, it seems not reasonable to have both of them at the same time. 
image

maybe consider:
1. deactive any other `deprecated enviroment` filter when a new  deprecated enviroment filter is chosen, or
2. use customized check box to filter the deprecated enviroment, which seems to be more user friendly.

DnsZhou (changes requested):
also see following comments.

DnsZhou (approved):
update verified from my local and lgtm.

PR #525, opened 2024-05-28
[NCL-8708] Fix bulk shift state persistence after changing browser tabs

patrikk0123:


DnsZhou (approved):
I verified the changes on my local, seems there isn't any impact on other tabs now.

PR #534, opened 2024-06-04
NCL-8725 Use standard Build Config endpoint

matedo1:


DnsZhou (commented):
Sorry for the late review, the code looks good and I verified it from my local. no error or warning related to this parts.

PR #535, opened 2024-06-05
[NCL-8496] Local storage unification

patrikk0123:


DnsZhou:
Hi @patrikk0123 , feel free to resolve the conflicts first, and I will have a code review on the code. Thanks.

src/components/TopBar/TopBar.tsx
```diff
@@ -36,28 +36,29 @@ interface ITopBarProps {
 export const TopBar = ({ children, id, type, icon, hideCloseButton = false }: React.PropsWithChildren<ITopBarProps>) => {
   const [isOpen, setIsOpen] = useState<boolean>(true);
 
-  const statusLocalStorageId = useMemo(() => `${id}-status`, [id]);
-  const textLocalStorageId = useMemo(() => `${id}-text`, [id]);
+  const statusLocalStorageKey = useMemo(() => `${id}-status`, [id]);
+  const textLocalStorageKey = useMemo(() => `${id}-text`, [id]);
 
+  // TODO: change to local storage?
```
DnsZhou, line 42: Personally, I think keeping session storage is a good choice for the top bar, specifically so that users will see the warning again when they open a new tab or visit the UI again.  If we change to local storage, then users will not be able to see the top bar any more; they may miss some important message just because they hide the top bar once. I would suggest we have another hook like `useSessionStorage` for this case.

DnsZhou: @matedo1 what do you think?

DnsZhou (approved):
Code mostly looks good to me and verified from my local env with stage backend. Just add the TSDocs and maybe consider my comments about the session storage parts.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type TLocalStorageValue = string | boolean | number;
+
+interface IUseLocalStorageProps<T extends TLocalStorageValue> {
+  localStorageKey: string;
+  initialValue: T;
+  localStorageValueParser?: (localStorageValue: string) => T;
+}
+
+export const useLocalStorage = <T extends TLocalStorageValue>({
```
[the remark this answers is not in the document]
matedo1: Optionally this could be made more general to include `sessionStorage`, etc. -> `useStorage` accepting additional parameter defining which storage should be used.

An alternative solution might involve creating separate hooks like `useSessionStorage`, etc. but this approach could lead to unnecessary code duplication.

src/common/constants.ts
```diff
@@ -92,5 +92,9 @@ export const MESSAGE_WAIT_AND_REFRESH = 'Wait 5 minutes and try to refresh your
 // Storage keys (like localStorage), see NCL-8496
 export const StorageKeys = {
```
matedo1, line 93: Optionally this could be moved to the `hooks/useLocalStorage.ts`

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
+import { useState } from 'react';
+
+type TLocalStorageValue = string | boolean | number;
+
+interface IUseLocalStorageProps<T extends TLocalStorageValue> {
+  localStorageKey: string;
+  initialValue: T;
+  localStorageValueParser?: (localStorageValue: string) => T;
```
matedo1, line 8: If `initialValue` is mandatory, then `localStorageValueParser` could be theoretically determined from it = if `initialValue` is `boolean`, then `booleanLocalStorageValueParser` could be used automatically. 

But I am not sure about possible unwanted side-effects as it can make this hook less reusable and less explicit.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
+import { useState } from 'react';
+
+type TLocalStorageValue = string | boolean | number;
+
+interface IUseLocalStorageProps<T extends TLocalStorageValue> {
+  localStorageKey: string;
```
matedo1, line 6: Just `key` or `storageKey` would be probably enough.

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (stringifiedValue) {
+      localStorage.setItem(localStorageKey, stringifiedValue);
+    } else {
+      localStorage.removeItem(localStorageKey);
+    }
+
+    setStoredValue(value);
+  };
+
+  return { storedValue, storeToLocalStorage };
```
matedo1, line 32: Just `storeToStorage` would be probably enough, `storeValue` could be also an option, but it's very similar to `storedValue`.

matedo1 (commented):
Just minor optional comments, mostly about making it more generalised. Thanks!

src/components/TopBar/TopBar.tsx
patrikk0123: I generalized the hook into `useStorage`

src/hooks/useStorage.ts
```diff
@@ -0,0 +1,66 @@
...
+export const useStorage = <T extends TStorageValue>({
+  storageKey,
+  initialValue,
+  storage = localStorage,
+}: IUseStorageProps<T>) => {
+  const storageValueParser = useCallback(
+    (rawStorageValue: string): T => {
+      if (typeof initialValue === 'number') {
+        return Number(rawStorageValue) as T;
+      } else if (typeof initialValue === 'boolean') {
```
matedo1, line 30: `else` could be skipped

src/hooks/useStorage.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+    [initialValue]
+  );
+
+  const [storageValue, setStorageValue] = useState<T>(() => {
+    const rawStorageValue = storage.getItem(storageKey);
+    return rawStorageValue ? storageValueParser(rawStorageValue) : initialValue;
+  });
+
+  const storeToStorage = (value: T) => {
```
matedo1, line 44: `useCallback()` is not used in this case, @patrikk0123 and @DnsZhou didn't you notice some unwanted re-renders?

matedo1:
Just minor comments, I will leave the final testing and approval up to @DnsZhou

DnsZhou (approved):
Changes look good to me now, feature verified from my local and no unnecessary rerendering was found with the implementation of  `useCallback`.

src/hooks/useStorage.ts
patrikk0123: I encapsulated in the useCallback now too

PR #536, opened 2024-06-10
[NCL-8709] Do not fetch Alignment log of Build in progress

patrikk0123:


DnsZhou (approved):
lgtm and verified from my local.

PR #539, opened 2024-06-12
[NCL-7537] Upgrade PatternFly from v4 to v5

patrikk0123:


src/components/BuildsList/BuildsList.tsx
```diff
@@ -181,7 +181,7 @@ export const BuildsList = ({ serviceContainerBuilds, columns = defaultColumns, c
               id={StorageKeys.isBuildsListCompactMode}
               label="Compact Mode"
               isChecked={isCompactMode}
-              onChange={(checked) => {
+              onChange={(_, checked) => {
```
jomrazek, line 184: Out of curiosity, are you saying by "_" that there is no event?

jomrazek (approved):
Looks ok to my knowledge

matedo1 (approved):
Looks good, I added a few more details to the jira comment

src/components/BuildsList/BuildsList.tsx
patrikk0123: the event is there, `_` or string prepended by `_`, is a convention to name variables to denote that the value is not being used (linter if properly set complains the variable is declared but unused)

src/components/AdministrationPage/AdministrationPage.tsx
```diff
@@ -172,6 +156,7 @@ export const AdministrationPage = () => {
                     autoComplete="off"
                     {...register<string>(pncStatusEntityAttributes.banner.id)}
                   />
+                  <FormInputHelperText variant="error">{getFieldErrors(pncStatusEntityAttributes.banner.id)}</FormInputHelperText>
```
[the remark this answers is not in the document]
patrikk0123: PF `FormHelperText` component does not require it nor has any prop for it (has no `isHidden`)
if the children of that component are empty (including an empty string) the helper is not displayed

I included the `isHidden` prop on `FormInputHelperText` when the helper string is not strictly empty but has to be disabled by other logic

PR #544, opened 2024-06-27
[NCL-7537] Upgrade deprecated Dropdown on BuildStartButton

patrikk0123:


DnsZhou (approved):
Component verified from my local, both compact mode and normal was tested without any problem. Code looks good to me

PR #545, opened 2024-07-01
[NCL-7537] Deprecated Select component upgrade

patrikk0123:


src/components/Select/Select.tsx
```diff
@@ -0,0 +1,69 @@
...
+  id,
+  className,
+  isOpen,
+  onToggle,
+  value,
+  onChange,
+  validated,
+  placeholder,
+  onBlur,
+  isToggleFullWidth = true,
```
DnsZhou, line 32: I understand there are 3 components we are using full width, but is there any other reason this default param needs to be altered to `true`? The default `isFullWidth` param of `MenuToggleProps` is `false`, and I can see you have to set `isToggleFullWidth=false` back in 2 components as well. I would suggest keeping the same default param logic with PF version components.

DnsZhou (changes requested):
Verified from my local, just some minor changes.

src/components/Select/Select.tsx
patrikk0123: The component will be used mostly as form select where the full width is needed.

DnsZhou (approved):

PR #546, opened 2024-07-02
[NCL-8383] Fix: wrong SCM repo entity attributes on SCM create/edit page

patrikk0123:


DnsZhou (changes requested):
Hi @patrikk0123 , the fix mostly looks good to me, just one display issue when switching between different alert status:
If there is a blue `This external repository is already synced.` alert displaying(usually happens when changing a URL back to the original one), and modifying it to some other url, the alert turns red shortly while loading, then it turns to blue as expected, but there was no problem with the scm at all.
see jira for screenshots.

PR #554, opened 2024-07-23
[NCL-8599] Handle uncaught cancelled HTTP request errors

patrikk0123:


DnsZhou (approved):
Verified from my local that the cancelled request error did not happen any more, cancelled requests are properly handled on all pages. Code looks good from my side, all serviceContainers have been updated properly.

PR #556, opened 2024-07-24
[NCL-8765] Integrate getComponentQueryParamsObject hook into useQueryParamsEffect effect

patrikk0123:


matedo1 (changes requested):
When accessing Build Config revision tab, there is never ending "Uncaught runtime errors" loop

matedo1 (approved):

PR #558, opened 2024-08-01
Share state of user authentication - broadcast

jomrazek:


src/services/keycloakService.ts
```diff
@@ -98,8 +99,11 @@ class KeycloakService {
    */
   public isAuthenticated(): boolean {
     this.checkKeycloakAvailability();
+    const authenticated = this.keycloakAuth.authenticated!;
 
-    return this.keycloakAuth.authenticated!;
+    authBroadcastService.send(authenticated);
```
patrikk0123, line 104: I suppose this might cause too much re-render of the `AppLayout` each time `isAuthenticated` is queried, which is not scarce.

If this would be achievable in the `AppLayout`:
```js
useEffect(() => {
    // update the user state
  }, [location.pathname]);
```

patrikk0123: Context API is probably not usable here, because it cannot be connected to the Keycloak code, if the Keyclode JS library does not provide event listeners.

package.json
```diff
@@ -14,6 +14,7 @@
     "@types/react": "17.0.76",
     "@types/react-dom": "^17.0.11",
     "axios": "^1.2.6",
+    "broadcast-channel": "^7.0.0",
```
matedo1, line 17: Let's prefer standard javascript libraries instead of external ones.

src/services/authBroadcastService.ts
```diff
@@ -0,0 +1,20 @@
+export interface IAuthBroadcastMessage {
+  authenticated: boolean;
+}
+
+class AuthBroadcastService {
```
matedo1, line 5: Consider using pure function instead of class, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/main/src/services/webSocketClient.ts

src/services/keycloakService.ts
matedo1: Broadcasting sounds like a preferable option, as one of its main advantages is that it can be used to detect user actions in other tabs within a same origin.

Maybe some nice feature could be to send user object directly instead of `authenticated` (boolean), so that whole user object would be directly automatically accessible via argument everywhere. When user is not available, `null` could be passed instead.

src/AppLayout.tsx
```diff
@@ -54,7 +55,11 @@ import pncLogoText from './pnc-logo-text.svg';
 export const AppLayout = () => {
   const webConfig = webConfigService.getWebConfig();
 
-  const user = keycloakService.isKeycloakAvailable ? keycloakService.getUser() : null;
+  const [user, setUser] = useState(keycloakService.isKeycloakAvailable ? keycloakService.getUser() : null);
+
+  authBroadcastService.getChannel().onmessage = (event: MessageEvent<IAuthBroadcastMessage>) => {
```
matedo1, line 60: I think some technical details could be encapsulated a little bit. 

`authBroadcastService.getChannel().onmessage` can be converted to something like:

1) `authBroadcastService.onMessage((event) => { setUser(..) });` 

or even better:

2) `authBroadcastService.addMessageListener((event) => { setUser(..) });` which provides better options when some additional global actions like message parsing etc. are required.

matedo1 (changes requested):

matedo1 (changes requested):

matedo1 (approved):

PR #559, opened 2024-08-01
Share state of user authentication - local storage

jomrazek:


matedo1 (commented):
Let's go with https://github.com/project-ncl/pnc-web-ui-react/pull/558 instead.

PR #569, opened 2024-09-04
[NCL-8740] Make SCM repo creation alert displayed also when creation is WIP

patrikk0123:


DnsZhou (commented):
Hi @patrikk0123 , the change looks good, but I believe this jira (8740) also requires for changes on creating build config page, are you going to do it in a separate PR? I think they can be done in the same jira as there isn't too much change.

patrikk0123:
@DnsZhou 
I will do it in the same PR. Didn't thought of it before.

DnsZhou (approved):
looks good to me

PR #577, opened 2024-09-24
[NCL-8834] Fix Log Viewer scrolling

patrikk0123:

PR #578, opened 2024-09-25
[NCL-8838] Change useDataBuffer hook to output all available data soon as possible

patrikk0123:


patrikk0123:
Depends on: https://github.com/project-ncl/pnc-web-ui-react/pull/577

matedo1:
Hi @patrikk0123 , let's wait for #577 to be merged first, I would like to improve this PR a little bit so that `setDataOut` is performed only when necessary, as there is currently an infinite loop. Here's a basic idea:

```ts
const throttleCallback = throttle(() => {
    console.log('useDataBuffer -> setInterval -> setDataOut');
    setDataOut(dataIn.current);
  }, 1000);

  const addLines = useCallback(
    (lines: string[]) => {
      dataIn.current = [...dataIn.current, ...preprocessor(lines)];

      console.log('useDataBuffer -> addLines');

      throttleCallback();
    },
    [preprocessor]
  );
```

matedo1 (approved):

PR #584, opened 2024-10-08
[NCL-8838] Improve throttle setting to fix first-line log misbehaviour

DnsZhou:
Also, remove Resume footer button for static logs

### DIFF
--- a/src/components/LogViewer/LogViewer.tsx
+++ b/src/components/LogViewer/LogViewer.tsx
@@ -162,7 +162,7 @@ export const LogViewer = ({ isStatic = false, data, customActions }: ILogViewerP
       data={renderedData}
       onScroll={onScroll}
       toolbar={<HeaderToolbar />}
-      footer={isPaused && !isFollowing && <FooterButton />}
+      footer={!isStatic && isPaused && !isFollowing && <FooterButton />}
       isTextWrapped={areLinesWrapped}
     />
   );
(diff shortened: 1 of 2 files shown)


src/hooks/useDataBuffer.ts
```diff
@@ -29,9 +29,13 @@ export const useDataBuffer = (
...
     () =>
-      throttle(() => {
-        setDataOut(dataIn.current);
-      }, delay),
+      throttle(
+        () => {
+          setDataOut(dataIn.current);
+        },
+        delay,
+        { leading: false, trailing: true }
```
DnsZhou, line 37: The root cause of the misbehaviour of the first line is from lodash that was introduced in useDataBuffer.
when leading set to true, the first call is executed immediately (which we do not want).
when trailing is set to true, the subsequent calls within the throttle delay are ignored until the delay period elapses.

src/components/LogViewer/LogViewer.tsx
```diff
@@ -162,7 +162,7 @@ export const LogViewer = ({ isStatic = false, data, customActions }: ILogViewerP
       data={renderedData}
       onScroll={onScroll}
       toolbar={<HeaderToolbar />}
-      footer={isPaused && !isFollowing && <FooterButton />}
+      footer={!isStatic && isPaused && !isFollowing && <FooterButton />}
```
DnsZhou, line 165: when scrolling up for a static log(like finished deliverable analysis), there is still a resume button, which I think should be removed.

patrikk0123 (approved):

matedo1 (approved):

src/hooks/useDataBuffer.ts
matedo1: Commit title was not changed

PR #588, opened 2024-10-10
[NCL-8851] Split Log Viewer toolbar and the log body into separate components

patrikk0123:


src/components/LogViewer/LogViewerBase.tsx
```diff
@@ -0,0 +1,91 @@
...
+  areLinesWrapped: boolean;
+  toolbar?: ReactNode;
+}
+
+interface IOnScrollProps {
+  scrollOffsetToBottom: number;
+  scrollUpdateWasRequested: boolean;
+}
+
+export const LogViewerBase = ({
```
DnsZhou, line 25: Maybe consider using `React.memo` for child component, but I will leave it to you to decide whether to use memo or not.

src/components/LogViewer/LogViewerToolbar.tsx
```diff
@@ -0,0 +1,95 @@
...
+  isStatic: boolean;
+  setIsPaused: (value: boolean) => void;
+  isFollowing: boolean;
+  setIsFollowing: (value: boolean) => void;
+  areLinesWrapped: boolean;
+  setAreLinesWrapped: (value: boolean) => void;
+  customActions?: ReactNode[];
+}
+
+export const LogViewerToolbar = ({
```
DnsZhou, line 18: Maybe consider using `React.memo` for child component, but I will leave it to you to decide whether to use memo or not.

DnsZhou (approved):
The split component looks good to me and verified from my local. I was just thinking maybe we can use memo for child components(see https://react.dev/reference/react/memo), but I will leave it to you to decide whether to use memo or not.

src/components/LogViewer/LogViewerBase.tsx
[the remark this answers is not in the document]
patrikk0123: i don't find that necessary, it's only internal component for the log viewer

src/components/LogViewer/LogViewerToolbar.tsx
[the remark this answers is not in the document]
patrikk0123: ditto

src/components/LogViewer/LogViewerBase.tsx
patrikk0123: I think it is a good idea, I encapsulated it in the memo

PR #589, opened 2024-10-10
[NCL-8847] Add search bar for live log viewer

DnsZhou:


DnsZhou:
will rebase after https://github.com/project-ncl/pnc-web-ui-react/pull/588 merged as this pr has more changes.

src/components/LogViewer/LogViewer.tsx
```diff
@@ -102,6 +105,23 @@ export const LogViewer = ({ isStatic = false, data, heightOffset = 0, customActi
     return () => clearTimeout(timeoutId);
   }, []);
 
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === '/') {
+        event.preventDefault();
+        const searchInput = document.querySelector<HTMLInputElement>(
```
matedo1, line 112: Did you consider using `ref={ref}` attribute?

```ts
const ref: React.MutableRefObject<HTMLInputElement | null> = React.useRef(null);
```

src/components/LogViewer/LogViewer.tsx
```diff
@@ -11,9 +11,10 @@ import styles from './LogViewer.module.css';
 const HEIGHT_DEFAULT_OFFSET = 300;
 
 interface ILogViewerProps {
-  isStatic?: boolean;
   data: string | string[];
+  isStatic?: boolean;
   heightOffset?: number;
+  showSearchBar?: boolean;
```
matedo1, line 17: I would avoid this property, we should display it always for all logs.

matedo1 (changes requested):
On pages that contain only log, like `Build Log`, `Alignment Log`, etc. I would like to test autofocus right after the page is loaded.

src/components/LogViewer/LogViewer.tsx
DnsZhou: I tried but we can't use ref to access the internal input element of LogViewerSearch because LogViewerSearch from PF does not support ref forwarding or provide a way to pass a ref to the internal SearchInput component.

matedo1: I've checked the implementation details and I think it would be passed automatically.

DnsZhou: After getting deep into the PF search bar components, the SearchInput component and its underlying components do not forward the ref all the way down to the actual <input> element.

Even though SearchInput accepts a ref prop and passes it as innerRef to SearchInputBase, and then to TextInputGroupMain, the chain breaks because TextInputGroupMain does not pass innerRef to the underlying <input>. See the following code from PF: https://github.com/patternfly/patternfly-react/blob/a3ffb39a0cc2c9130f7db86ad3186787ba12648e/packages/react-core/src/components/SearchInput/SearchInput.tsx#L292-L305

So I believe we have to use `document.querySelector` under this circumstance. @matedo1

src/components/LogViewer/LogViewerToolbar.tsx
```diff
@@ -24,74 +26,120 @@ export const LogViewerToolbar = memo(
...
+      return () => {
+        window.removeEventListener('keydown', handleKeyDown);
+      };
+    }, []);
+
+    useEffect(() => {
+      if (autofocusSearchBar) {
+        const timeoutId = setTimeout(() => {
+          const searchInput = document.querySelector<HTMLInputElement>(
+            '.pf-v5-c-log-viewer__header .pf-v5-c-text-input-group__text-input'
```
matedo1, line 61: Did you consider moving searchInput (or some part of it) into the separated constant to prevent code duplication?

src/components/LogViewer/LogViewerToolbar.tsx
```diff
@@ -24,74 +26,120 @@ export const LogViewerToolbar = memo(
...
-            customActions.map((node, index) => (
-              <Fragment key={index}>
+  }: ILogViewerToolbarProps) => {
+    useEffect(() => {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        const target = event.target as HTMLElement;
+        const tagName = target.tagName.toLowerCase();
+        const isInput = tagName === 'input' || tagName === 'textarea' || target.isContentEditable;
+
+        if (isInput) {
```
matedo1, line 38: Please add comment.

matedo1 (changes requested):

src/components/LogViewer/LogViewerToolbar.tsx
```diff
@@ -24,74 +26,120 @@ export const LogViewerToolbar = memo(
...
+        return () => clearTimeout(timeoutId);
+      }
+    }, [autofocusSearchBar]);
+
+    return (
+      <Toolbar>
+        <ToolbarContent alignItems="center">
+          <ToolbarGroup>
+            <ToolbarItem>
+              <LogViewerSearch minSearchChars={1} placeholder="Type / to search" />
```
patrikk0123, line 75: Ideally separate the search bar and all related logic outside from the toolbar

DnsZhou: Just like Top and Bottom buttons, the search bar is highly bonded to the toolbar of the log viewer only, and search bar will need to be used for every log viewer(as Martin mentioned earlier in this PR), so I think we should keep it in toolbar.

DnsZhou: @patrikk0123 updated

matedo1 (approved):
Just minor comment, thanks.

src/components/LogViewer/LogViewer.tsx
```diff
@@ -10,9 +10,10 @@ import styles from './LogViewer.module.css';
 
 interface ILogViewerProps {
   isStatic?: boolean;
-  data: string | string[];
   heightOffset?: number;
+  autofocusSearchBar?: boolean;
   customActions?: ReactNode[];
+  data: string | string[];
```
patrikk0123, line 16: please do not change the line position of `data` attribute in the interface definition

src/components/LogViewer/LogViewer.tsx
```diff
@@ -35,9 +36,17 @@ interface ILogViewerProps {
  *
  * @param isStatic - true for static variant if whole log is available at once, false for live variant for dynamically loaded log lines
  * @param heightOffset - offset for the responsive height for the LogViewer component
+ * @param autofocusSearchBar - If true, activates the search input when the component mounts
+ * @param customActions - An array of React nodes that will be rendered as additional actions within the toolbar.
```
patrikk0123, line 40: dot at the end of the line => either put it on each line on nowhere, thanks

src/components/LogViewer/LogViewerSearchBar.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { memo, useEffect } from 'react';
+
+interface ILogViewerSearchBarProps {
+  autofocusSearchBar: boolean;
+}
+
+export const LogViewerSearchBar = memo(({ autofocusSearchBar }: ILogViewerSearchBarProps) => {
+  const focusSearchInput = () => {
+    const searchInput = document.querySelector<HTMLInputElement>(
+      '.pf-v5-c-log-viewer__header .pf-v5-c-text-input-group__text-input'
```
patrikk0123, line 11: maybe put the string into constant outside the compoennt (same file)

patrikk0123 (commented):
I tested it locally and it works nicely

I miss the navigation through search matches with ENTER or so, but it should have been done by PatternFly itself

PR #592, opened 2024-10-17
[NCL-8893] Split live from static build log

patrikk0123:


src/components/BuildPages/BuildPages.tsx
```diff
@@ -127,14 +129,19 @@ export const BuildPages = () => {
 
   const isLogged = !serviceContainerBuild.data?.status || isBuildWithLog(serviceContainerBuild.data.status);
   const isLoggedTooltip = !isLogged ? `Builds with status ${serviceContainerBuild.data!.status} are not logged.` : '';
+  const liveLogLinkTooltip = !isBuilding ? `Not available while not building.` : '';
+  const staticLogLinkTooltip = isBuilding ? `Not available while building.` : '';
```
matedo1, line 133: I would go with the same message which is already displayed on details tab: `Build is not finished yet`

image

src/components/BuildPages/BuildPages.tsx
```diff
@@ -127,14 +129,19 @@ export const BuildPages = () => {
 
   const isLogged = !serviceContainerBuild.data?.status || isBuildWithLog(serviceContainerBuild.data.status);
   const isLoggedTooltip = !isLogged ? `Builds with status ${serviceContainerBuild.data!.status} are not logged.` : '';
+  const liveLogLinkTooltip = !isBuilding ? `Not available while not building.` : '';
```
matedo1, line 132: Here we could use some equivalent phrasing, something like `Build is not in progress` to avoid double `not`.

src/components/BuildPages/BuildPages.tsx
```diff
@@ -127,14 +129,19 @@ export const BuildPages = () => {
...
+  const staticLogLinkTooltip = isBuilding ? `Not available while building.` : '';
 
   const pageTabs = (
     <PageTabs>
       <PageTabsItem url="details">Details</PageTabsItem>
-      <PageTabsItem url="build-log" isDisabled={!isLogged} tooltip={isLoggedTooltip}>
+      <PageTabsItem url="live-log" isDisabled={!isLogged || !isBuilding} tooltip={isLoggedTooltip || liveLogLinkTooltip}>
+        Live Log
+      </PageTabsItem>
+      <PageTabsItem url="build-log" isDisabled={!isLogged || isBuilding} tooltip={isLoggedTooltip || staticLogLinkTooltip}>
```
matedo1, line 141: I've notices if there are two disabled tab items next to each other, then they are visually merged. Unfortunately adding `border-right: 2px solid white` does not seem like a good solution because it breaks bottom shadow a little bit. You can try to fix it, or we can fix it in separated PR later.

matedo1, line 141: We can consider disabling also some other tabs items like Artifacts, Dependencies and Brew Push

matedo1 (changes requested):
Once a build is finished, I've noticed the live log page is refreshed (live log is reprinted again from scratch) and no tab item is being active (because I am on Live Log tab which is actually disabled). Would it be possible to redirect to Build Log tab instead?

src/components/BuildPages/BuildPages.tsx
patrikk0123: pretty much way PF has designed it
https://www.patternfly.org/components/tabs/#default-tabs

matedo1: Ok, let's keep for now, thanks

src/components/BuildPages/BuildPages.tsx
```diff
@@ -127,32 +129,41 @@ export const BuildPages = () => {
 
   const isLogged = !serviceContainerBuild.data?.status || isBuildWithLog(serviceContainerBuild.data.status);
   const isLoggedTooltip = !isLogged ? `Builds with status ${serviceContainerBuild.data!.status} are not logged.` : '';
+  const liveLogLinkTooltip = !isBuilding ? `Build is in not progress.` : '';
```
matedo1, line 132: There is a typo, `Build is not in progress`.

matedo1 (approved):
Just minor comment. Thanks

PR #594, opened 2024-10-23
[NCl-8897] Remove trailing \n from bifrost log lines

patrikk0123:


matedo1:
Hi @DnsZhou , is it proper solution when you compare it with AngularJS based UI?

DnsZhou:
In the old Bifrost Js library, we are not using PatternFly LogViewer at all. 

Each message received in the WebSocket callback (this.client.subscribe(...)) was treated as an individual line, and it was appended using [logViewer.appendLine() ](https://github.com/project-ncl/bifrost-js/blob/1dba13b3146ef210e5accd5d05835785413ede9e/packages/web-component/src/BifrostLogViewer.ts#L50) that is defined here:https://github.com/project-ncl/bifrost-js/blob/1dba13b3146ef210e5accd5d05835785413ede9e/packages/web-component/src/LogViewer.ts#L106
`this.buffer.queueAppend(this.createLineElement(text));`
Also we used [createLineElement](https://github.com/project-ncl/bifrost-js/blob/1dba13b3146ef210e5accd5d05835785413ede9e/packages/web-component/src/LogViewer.ts#L142-L154) to create each line individually.  

Given that situation, I would say removing /n manually at the end of the message can be a reasonable solution. But maybe we should only remove the \r\n from the end of the message? not sure if some message have \r\n on purpose that we may need to retain? @matedo1 @patrikk0123

matedo1 (approved):

PR #598, opened 2024-10-30
[NCL-8930] Fix: Build History panel is reloaded and reseted for each tab

patrikk0123:


src/components/BuildConfigPages/BuildConfigPages.tsx
```diff
@@ -119,7 +121,7 @@ export const BuildConfigPages = ({ componentIdBuildHistory = 'bh1' }: IBuildConf
   );
 
   const pageTabs = (
-    <PageTabs>
+    <PageTabs queryParams={componentQueryParamsString}>
       <PageTabsItem url="details">Details</PageTabsItem>
```
matedo1, line 125: Did you try to pass query parameters directly to individual PageTabsItem to avoid all provider related logic?

`<PageTabsItem url={appendQueryParams('details', buildHistoryQueryParamsString)}>Details</PageTabsItem>`

or 

`<PageTabsItem url={appendBuildHistoryQueryParams('details')}>Details</PageTabsItem>`

it seems like a more straightforward approach that would probably do the same job easily AND it allows you to apply different params for the specific tabs if it's required on some other tab based pages.

patrikk0123: I wanted to avoid defining query params for each tab item, hence the context api logic and all.

matedo1: I will leave it up to you. Possibly @DnsZhou, do you have any preferences?

matedo1 (approved):

src/components/BuildConfigPages/BuildConfigPages.tsx
DnsZhou: In the current use case, PageTabsItem and PageTabs are indeed tightly coupled, so passing parameters at the PageTabs level can reduce the repetitive work of setting parameters for each PageTabsItem. 

However, I agree with Martin's perspective: defining query parameters directly at the PageTabsItem level appears more straightforward and clear. If we need to implement more dynamic requirements in the future—such as individual tabs needing different parameters or passing new parameters in specific scenarios—we would still need to pass parameters at the PageTabsItem level.

Anyway in the latest code I can see it was already changed to use PageTabsItem level param, and sorry for the late comments!

PR #599, opened 2024-10-31
[NCL-8826] Clear Brew Push legacy log when trigger new Brew Push

DnsZhou:


matedo1:
I would like to try a little bit different approach, especially `setResetLogTrigger()` related logic should be avoided. What about implement `addOnOpenListener()` and clear, or just force page reload when new WebSocket channel is open? Let's discuss additional details on PM.

DnsZhou:
@matedo1  We first need to know when the new WS channel should be opened, this is a chicken-egg problem. 

New ws channel opens with the initialization of useBifrostWebSocketEffect when we load buildBrewPushPage(and other pages with log viewer that contains useBifrostWebSocketEffect).

We anyway need to pass a message to buildBrewPush page so that it knows a new brew push is triggered, then, either clear the buffer or start a new ws channel will both do the job. I think clear the buffer is enough.

matedo1:
I understand your worries, but log should ideally reseted always once new WebSocket is open = once old WebSocket connection is close. I've just tested it locally and it seems it's working well. I slightly improved it and I am using `addOnCloseListener()` listeners instead, then log reset is performed (page refresh is problematic, let's avoid it).

DnsZhou:
Hi @matedo1 , feel free to review this version. I also fixed the empty web socket problem for brew push log(this existed before my change for clearing the log, it was caused by initialing ws too early before brew push id was on hand )

src/components/BuildBrewPushPage/BuildBrewPushPage.tsx
```diff
@@ -69,19 +69,26 @@ export const BuildBrewPushPage = () => {
...
-      matchFilters: `${brewPushLogMatchFiltersPrefix1}${buildId}${brewPushLogMatchFiltersPrefix2}${serviceContainerBrewPush.data?.id}`,
+      matchFilters: `${brewPushLogMatchFiltersPrefix1}${buildId}${brewPushLogMatchFiltersPrefix2}${
+        serviceContainerBrewPush.data?.id ?? ''
+      }`,
     }),
     [buildId, serviceContainerBrewPush.data?.id]
   );
 
+  const preventListening = useMemo(() => !serviceContainerBrewPush.data?.id, [serviceContainerBrewPush.data?.id]);
+  const handleOnClose = useCallback(() => {
```
matedo1, line 80: Is it needed to use `useCallback` again? I think `resetLogBuffer` is already memoized.

src/hooks/useBifrostWebSocketEffect.ts
```diff
@@ -54,6 +54,7 @@ interface IUseBifrostWebSocketEffectOptions extends IUsePncWebSocketEffectOption
     prefixFilters: string;
     matchFilters: string;
   };
+  onClose?: () => void;
```
matedo1, line 57: Please add some TSDoc

matedo1 (changes requested):

matedo1 (approved):

PR #600, opened 2024-11-11
[NCL-8899] Add announcement redirect to old UI when React UI is set as primary UI

DnsZhou:


src/components/TopBar/OldUIAnnouncement.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return isClose ? null : (
+    <Alert
+      id="old-ui-announcement-bar"
+      variant={AlertVariant.custom}
+      customIcon={<ApplicationsIcon />}
+      actionClose={<AlertActionCloseButton onClose={closeAnnouncement} />}
+      title={
+        <>
+          You are currently using PNC new Web UI. If you prefer the old version, you can{' '}
```
matedo1, line 31: Welcome to the new PNC Web UI. The [old UI version](http://example.com) is temporarily available, but it will soon be decommissioned.

src/components/TopBar/OldUIAnnouncement.tsx
```diff
@@ -0,0 +1,41 @@
+import { Alert, AlertActionCloseButton, AlertVariant } from '@patternfly/react-core';
+import { ApplicationsIcon, ArrowIcon } from '@patternfly/react-icons';
+import { useState } from 'react';
+
+const PNC_OLD_UI_URL_TRIGGER = 'secondary';
+
+export const OldUIAnnouncement = () => {
+  const [isClose, setIsClose] = useState<boolean>(false);
```
matedo1, line 8: I think we should use local storage or something to keep the closed state more permanent.

src/components/TopBar/OldUIAnnouncement.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const oldUIUrl = process.env.REACT_APP_PNC_OLD_UI_WEB;
+
+  /* Only display the announcement when old ui is set to secondary */
+  const shouldDisplay = !!(oldUIUrl && oldUIUrl.includes(PNC_OLD_UI_URL_TRIGGER));
```
matedo1, line 12: Isn't better to use a condition that checks if the current browser URL does not contain 'secondary'? Then we could merge everything in advance and it would be automatically displayed once backend changes are done.

matedo1 (changes requested):

src/components/TopBar/OldUIAnnouncement.tsx
DnsZhou: If the user closes the announcement and we store the status in local storage, they will not be able to reopen it anymore. Do you think we can add an option in the config toggle of the menu bar for the user to get to the old UI?
image

matedo1: PM

matedo1 (approved):

PR #613, opened 2024-12-05
[NCL-8947] Disable restore button for last revision

DnsZhou:


src/components/BuildConfigDetailPage/BuilConfigRevisionPage.tsx
```diff
@@ -12,10 +12,28 @@ export const BuildConfigRevisionPage = () => {
...
   const serviceContainerBuildConfigRevision = useServiceContainer(buildConfigApi.getRevision);
   const serviceContainerBuildConfigRevisionRunner = serviceContainerBuildConfigRevision.run;
+  const serviceContainerBuildConfig = useServiceContainer(buildConfigApi.getBuildConfig);
+  const serviceContainerBuildConfigRunner = serviceContainerBuildConfig.run;
+
+  const isCurrentRevision = () => {
+    if (!serviceContainerBuildConfig.data || !serviceContainerBuildConfigRevision.data) {
+      return false;
+    }
+    return serviceContainerBuildConfig.data?.modificationTime === serviceContainerBuildConfigRevision.data?.modificationTime;
```
matedo1, line 22: I have two points:

1) Would it be possible to use ID based comparison instead of `modificationTime` based? We have list of all revisions already loaded and we know the last one is the current one.

2) The list of all revisions is already loaded on `BuildConfigRevisionPages` and I think the last revision ID could be shared via `useOutletContext` ? So that no other http request is performed. Your current solution is loading new Build Config for each revision, let's try to avoid additional HTTP requests.

matedo1 (changes requested):

src/components/BuildConfigDetailPage/BuilConfigRevisionPage.tsx
DnsZhou: Finally I came out with a solution that looks good: using the url param to make sure only disable the restore button when pageindex is 1.

src/components/BuildConfigDetailPage/BuildConfigDetail.tsx
```diff
@@ -60,7 +62,7 @@ export const BuildConfigDetail = ({
             </ToolbarItem>
             {isRevisionVariant && (
               <ToolbarItem alignRight>
-                <BuildConfigRestoreModalButton toggleModal={toggleRestoreModal} />
+                <BuildConfigRestoreModalButton toggleModal={toggleRestoreModal} isDisabled={isCurrentRevision} />
```
matedo1, line 65: Please include a tooltip that explains the reason for being disabled.

src/components/BuildConfigRevisionPages/BuildConfigRevisionPages.tsx
```diff
@@ -36,13 +38,23 @@ export const BuildConfigRevisionPages = ({ componentId = 'r1' }: IBuildConfigDet
     { componentId, mandatoryQueryParams: listMandatoryQueryParams.pagination }
   );
 
+  useEffect(() => {
+    const pageIndex = new URLSearchParams(search).get('r1-pageIndex');
+    if (serviceContainerBuildConfigRevisions.data?.content?.length && pageIndex === '1') {
+      const latestRev = serviceContainerBuildConfigRevisions.data.content[0]?.rev;
+      setIsCurrentRevision(String(latestRev) === revisionId);
+    } else {
```
matedo1, line 46: Is `else` needed?

matedo1 (changes requested):

src/components/BuildConfigRevisionPages/BuildConfigRevisionPages.tsx
DnsZhou: I think if the else branch is omitted, when the condition is not met, the state of isCurrentRevision will not be updated and may retain its previous value when the component was dynamically updated.

src/components/BuildConfigRestoreModal/BuildConfigRestoreModalButton.tsx
```diff
@@ -4,12 +4,15 @@ import { ProtectedComponent } from 'components/ProtectedContent/ProtectedCompone
...
+  isDisabled?: boolean;
 }
 
-export const BuildConfigRestoreModalButton = ({ toggleModal }: IBuildConfigRestoreModalButtonProps) => (
+export const BuildConfigRestoreModalButton = ({ toggleModal, isDisabled = false }: IBuildConfigRestoreModalButtonProps) => (
   <ProtectedComponent>
-    <Button variant="primary" onClick={toggleModal} size="sm">
-      Restore
-    </Button>
+    <div title={isDisabled ? 'Restore button is disabled for the current revision' : undefined}>
```
matedo1, line 12: For disabled tooltips we use slightly different syntax, see for example https://github.com/project-ncl/pnc-web-ui-react/blob/eb23ce3ea54a65cbed9ac8bfe6c7f718171e52ae/src/components/BrewPushModal/BrewPushModalButton.tsx#L18

matedo1 (changes requested):

matedo1 (approved):

PR #626, opened 2025-02-25
[NCL-6996] Upgrade React to v18

DnsZhou:
- Updated React to version 18.3.1 and updated related dependencies.
 - Updated index.tsx and useServiceContainer.ts according to the new code changes.
 - Updated the test cases to comply with React V18 requirements.


src/hooks/useServiceContainer.ts
```diff
@@ -150,20 +149,18 @@ export const useServiceContainer = <T extends TServiceData, U extends TServicePa
       .then((response) => {
         // In a future React version (potentially in React 17) this could be removed as it will be default behavior
         // https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973
-        ReactDOM.unstable_batchedUpdates(() => {
```
matedo1, line 153: Did you test `ReactDOM.unstable_batchedUpdates` is not needed anymore?

src/hooks/useServiceContainer.ts
```diff
@@ -180,11 +177,9 @@ export const useServiceContainer = <T extends TServiceData, U extends TServicePa
         if (loadingCount.current <= 1) {
           // In a future React version (potentially in React 17) this could be removed as it will be default behavior
```
matedo1, line 178: Comments should be removed if `ReactDOM.unstable_batchedUpdates` is not used anymore.

matedo1 (changes requested):

src/hooks/useServiceContainer.ts
DnsZhou: Yes I verified it by using a test like this:

```  const [count, setCount] = useState(0);

  useEffect(() => {
    setTimeout(() => {
      setCount((c) => c + 1);
      setCount((c) => c + 1);
    }, 1000);
  }, []);

  console.log('Render count:', count);
```

also see article that mentioned this change: https://github.com/reactwg/react-18/discussions/21

matedo1 (dismissed):
Let's wait with merging until you confirm it's fully tested (see Jira for more information).

src/hooks/useResizeObserver.ts
DnsZhou, line 1: In React 18, due to changes in StrictMode and concurrent rendering, the timing and order of callback ref invocations may differ from before, causing the ResizeObserver not to trigger as expected and resulting in the state’s height remaining undefined.

src/components/BuildConfigCreateEditPage/BuildConfigCreateEditPage.tsx
DnsZhou, line 1: In React 18, because of changes in how rendering works, React now requires that every item in a list has a stable and unique key. Before, the key was set on the inner FormGroup, but the top-level element returned from the map was the FormInput component. Thus I have to move the key to the outer element (the FormInput).

src/components/BuildConfigsList/BuildConfigsList.tsx
DnsZhou, line 1: React 18 has enhanced validation for incorrect HTML structures, ensuring <Th> and <Td> elements are only placed inside <Tr>, <Thead>, <Tbody>, and <Table>. ProtectedComponent in a table cell caused additional <div> element and I think as a table header there is no need to wrap it with ProtectedComponent

src/components/BuildDependenciesPage/BuildDependenciesPage.tsx
DnsZhou, line 1: Due to enhanced validation for incorrect HTML structures, this need to be updated

matedo1:
Hi @patrikk0123 , this PR is reaching its final version, could you please briefly test it too?

src/components/ProductVersionMilestonesPage/ProductVersionMilestonesPage.tsx
```diff
@@ -57,7 +57,7 @@ export const ProductVersionMilestonesPage = ({ componentId = 'm1' }: IProductVer
           {latestProductMilestone && (
             <TextContent>
               <br />
-              <Text>
```
DnsZhou, line 60: I have to change it to div as `Text` will introduce <p> and no <div> allowed for its children. But PF Tooltip of ProductMilestoneReleaseLabel will anyway introduce div so have no other choices.

src/hooks/useResizeObserver.ts
```diff
@@ -9,26 +9,29 @@ import { useCallback, useEffect, useRef, useState } from 'react';
...
   const [height, setHeight] = useState<number>();
 
   const ref = useCallback((node: HTMLElement | null) => {
+    setNode(node);
+  }, []);
+
+  useLayoutEffect(() => {
     if (!node) return;
-    resizeObserver.current = new ResizeObserver(() => {
+    const observer = new ResizeObserver((entries) => {
```
DnsZhou, line 22: In React 18, due to changes in StrictMode and concurrent rendering, the timing and order of callback ref invocations may differ from before, causing the ResizeObserver not to trigger as expected and resulting in the state’s height remaining undefined.

src/components/BuildConfigCreateEditPage/BuildConfigCreateEditPage.tsx
```diff
@@ -910,10 +910,10 @@ export const BuildConfigCreateEditPage = ({ isEditPage = false }: IBuildConfigCr
 
               {Object.entries(buildParamData).map(([key, buildParam], index) => (
                 <FormInput<string>
+                  key={index}
```
DnsZhou, line 913: In React 18, because of changes in how rendering works, React now requires that every item in a list has a stable and unique key. Before, the key was set on the inner FormGroup, but the top-level element returned from the map was the FormInput component. Thus I have to move the key to the outer element (the FormInput).

src/components/BuildConfigsList/BuildConfigsList.tsx
```diff
@@ -137,9 +137,7 @@ export const BuildConfigsList = ({
                     </Th>
                   )}
                 {columns.includes(buildConfigEntityAttributes.actions.id) && (
-                  <ProtectedComponent>
-                    <Th width={15}>{buildConfigEntityAttributes.actions.title}</Th>
-                  </ProtectedComponent>
+                  <Th width={15}>{buildConfigEntityAttributes.actions.title}</Th>
```
DnsZhou, line 140: React 18 has enhanced validation for incorrect HTML structures, ensuring Th and Td elements are only placed inside Tr, Thead, Tbody, and Table. ProtectedComponent in a table cell caused additional div elements and I think as a table header there is no need to wrap it with ProtectedComponent

src/components/BuildDependenciesPage/BuildDependenciesPage.tsx
```diff
@@ -34,8 +34,8 @@ export const BuildDependenciesPage = ({ componentId = 'd1' }: IBuildDependencies
       <Toolbar>
         <ToolbarItem>
           <TextContent>
+            <Text component={TextVariants.h2}>Dependencies</Text>
```
DnsZhou, line 37: Due to enhanced validation for incorrect HTML structures, this was updated

src/hooks/useResizeObserver.ts
matedo1: @patrikk0123 , are you ok with related changes?

src/hooks/useResizeObserver.ts
```diff
@@ -9,26 +9,29 @@ import { useCallback, useEffect, useRef, useState } from 'react';
...
       window.requestAnimationFrame(() => {
-        setWidth(node.offsetWidth);
-        setHeight(node.offsetHeight);
+        entries.forEach((entry) => {
+          setWidth(entry.target.clientWidth);
+          setHeight(entry.target.clientHeight);
+        });
```
patrikk0123, line 27: for multiple entries, single width and height state is updated (there should be probably only one entry, but it looks weird)

patrikk0123 (approved):
I think the code looks good, but didn't test it locally

matedo1 (commented):
We are currently waiting for confirmation abou new WebSocket (bifrost) console log error, see Jira for more information

DnsZhou:
See the latest PR comments, the WS issue was moved to a new jira. 
On the other hand, I have checked the white screen bottom issue in NCL-8432 and it is not happening with my local.

Let me know if we need any other changes(especially about the useResizeObserver.ts changes) @matedo1

matedo1 (approved):

PR #641, opened 2025-06-09
[NCl-9058] Build Push and Milestone close

patrikk0123:


src/common/operationEntityAttributes.ts
```diff
@@ -0,0 +1,80 @@
+import { DeliverableAnalyzerOperation } from 'pnc-api-types-ts';
+
+import { TEntityAttributes } from 'common/entityAttributes';
+
+// this supertype is not coming from pnc-api-types-ts
```
DnsZhou, line 5: muti-line comments are suggested:
`/*
 * This supertype is not coming from pnc-api-types-ts,
 * but these props are the same for all Operations.
 */`

DnsZhou (commented):
Verified on my local dev env with BC: `TC30-temporary-push`, brew-push list component and details page looks all good and no error or warning related to this change.

The new "Close Milestone" modal looks good, and I was able to close a milestone with `Skip Brew push` enabled, but I can't close it without `Skip Brew push` option. Not sure if it is due to dev env endpoint issue? @patrikk0123 do you have any idea on that?

DnsZhou (approved):
other than is issue raised above, it looks good to me

matedo1 (approved):

PR #658, opened 2025-08-20
[NCl-8934] Update to PatternFly v6 fixes

patrikk0123:


src/components/DeliverableAnalysisDetailPage/DeliverableAnalysisDetailPage.tsx
```diff
@@ -178,19 +179,19 @@ export const DeliverableAnalysisDetailPage = ({ componentId = 'da1' }: IDelivera
...
+          <PageSectionHeader title={PageTitles.deliverableAnalysisLabelsHistory} />
+        </ToolbarItem>
+      </Toolbar>
+      <DeliverableAnalysisLabelsHistoryList
+        serviceContainerLabelsHistory={serviceContainerLabelsHistory}
+        componentId={componentId}
+      />
 
-      <LogViewerSection deliverableAnalysis={deliverableAnalysis} />
+      <PlainContentBox marginTop>
```
matedo1, line 192: [optional] In this case I think using a simple `<div class="m-t-global">` could be easier to maintain in the future as there can be needed multiple combinations and customizations. And there are for sure other similar places where margins are currently applied without `PlainContentBox` being used.

matedo1 (changes requested):

matedo1 (approved):

PR #659, opened 2025-08-25
[NCL-9225] Fix custom query param applied filter crashing application

patrikk0123:


src/utils/qParamHelper.ts
```diff
@@ -204,11 +204,7 @@ export const removeQParamItem = (id: string, value: TQParamValue, operator: IQPa
 
 /**
  * @param qParam - RSQL string: filename=like="%te%t%";status!=CANCELLED
- * @returns Object representing individual RSQL items deeply parsed:
- * {
- *   name: ['"%te%t%"'],
- *   status: ['CANCELLED']
- * }
+ * @returns Object representing individual RSQL items deeply parsed along with logical operators
```
matedo1, line 207: Can you please keep the return example that corresponds to `filename=like="%te%t%";status!=CANCELLED` ?

matedo1 (approved):

PR #665, opened 2025-09-08
[NCL-9104] Add more filter presets on Delivered Artifacts page

patrikk0123:


matedo1 (changes requested):
There are several browser console log errors that need to be fixed first, see the Jira ticket for more details.

PR #688, opened 2026-03-04
NCL-9487 Add PNC Web UI ability to authenticate to any OIDC server

patrikk0123:


src/components/AuthServiceStatusPage/AuthServiceStatusPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * This page can be displayed in two different contexts:
+ *  1) Status page:
+ *    a) Auth Service service is available
+ *    b) Auth Service service is NOT available
+ *  2) Error page - requested page (for example projects/create) could not be displayed due to Auth Service failure
+ */
+export const AuthServiceStatusPage = ({ errorPageTitle }: IAuthServiceStatusPageProps) => {
+  const auth = useAuth();
+
+  const content = <>TODO... If the error still persists please contact {MESSAGE_PNC_ADMIN_CONTACT}.</>;
```
jomrazek: Do we want TODO here?

jomrazek (dismissed):
I guess AuthServiceStatusPage is not in final form but other than that looks good to me.

src/services/authService.ts
```diff
@@ -0,0 +1,71 @@
...
+});
+
+userManager.getUser().then((user) => {
+  if (user && !user.expired) {
+    userService.fetchUser();
+  }
+});
+
+export enum AUTH_ROLE {
+  Admin = 'pnc-users-admin',
```
jomrazek: @matedo1 do we possibly want to have this in some config? Given upcoming changes to different SSO and possibly different roles? Just thinking...

jomrazek (dismissed):
Looks good.

src/services/authService.ts
patrikk0123: that's good idea

jomrazek: I think also 'Employee' role should have same approach. I'm not sure how this will be handled in new SSO and if default roles are same, but we should probably have both admin and user configurable or none.

patrikk0123: I added also User role

jomrazek (approved):
👌

PR #712, opened 2026-05-13
[NCL-9648] Apply shared workflows and Mend

patrikk0123:


.github/workflows/npm-ci.yml
```diff
@@ -0,0 +1,32 @@
+# https://docs.github.com/en/actions/guides/building-and-testing-nodejs
+
+name: Node.js CI with NPM
```
matedo1: We could use `"PR CI"`.

.github/workflows/npm-mend-pr.yml
```diff
@@ -0,0 +1,33 @@
+name: Mend CLI scan for NPM PR
```
matedo1: We could use `"PR Mend"`.

.github/workflows/npm-mend.yml
```diff
@@ -0,0 +1,36 @@
+name: Mend CLI scan for NPM
```
matedo1: We could use `"Cron Mend"`.

.github/workflows/npm-mend.yml
```diff
@@ -0,0 +1,36 @@
...
+  actions: read
+  checks: write
+  pull-requests: write
+  security-events: write
+
+jobs:
+  build:
+    uses: project-ncl/shared-github-actions/.github/workflows/npm-ci.yml@153b97777335953545bc8a907eb5f062af66b7f2 # v0.0.18
+    with:
+      test_command: npm test
```
matedo1, line 22: Let's add also `formatter_command`.

.github/workflows/npm-mend.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches:
+      - main
+  schedule:
+    - cron: '00 03 * * *'
+    - cron: '00 11 * * *'
+
+permissions:
+  contents: read
+  actions: read
+  checks: write
```
matedo1, line 14: I think `checks` and `pull-requests` permissions should not be required? WDYT @jomrazek

matedo1 (commented):
Thank you for this PR. Hi @jomrazek , feel free to check this PR too.

.github/workflows/npm-mend.yml
patrikk0123: can't be removed right now, because the reusable workflow declares these permissions. can be removed after these permissions are also removed from there.

PR #717, opened 2026-06-09
Remove old env vars

patrikk0123:


src/index.tsx
```diff
@@ -34,10 +34,6 @@ const App = () => {
   if (
     // https://example.com
     !pncUrl.startsWith(window.location.origin) &&
-    !(
-      import.meta.env.VITE_WEB_UI_URL === window.location.origin ||
-      import.meta.env.VITE_WEB_SECONDARY_UI_URL === window.location.origin
```
patrikk0123, line 39: I think these are no longer needed because UI runs on the same URL as REST API

patrikk0123:
I'd also remove `VITE_GERRIT_URL_BASE` and `VITE_GERRIT_STAGE_URL_BASE`,
but I can still see a few internal repos in Gerrit existing on devel, but not stage and prod.

`/pnc-web/scm-repositories?s1-sort=none&s1-pageIndex=1&s1-pageSize=50&s1-q=%28internalUrl%3Dlike%3D"%25code%25"%29`

`['code']: 'Gerrit'` seems to cover it

src/utils/urlParseHelper.ts
```diff
@@ -3,8 +3,7 @@ interface IScmRepositoryUrl {
 }
 
 const preDefinedScmsPrefix: { [key: string]: string } = {
-  [import.meta.env.VITE_GERRIT_URL_BASE || 'code']: 'Gerrit',
-  [import.meta.env.VITE_GERRIT_STAGE_URL_BASE || 'code']: 'Gerrit',
+  ['code']: 'Gerrit',
```
matedo1, line 6: Hi @jomrazek , can you please confirm this change is safe?

jomrazek: Yes that should be fine.

